PR #109, opened 2026-03-10
Migrate to Supabase: sync pipeline, server-side queries, remove static data

shaneweisz:
### Supabase sync pipeline
- Add Supabase schema (species table + taxa_summary materialized view) and sync pipeline scripts that fetch from IUCN Red List and GBIF APIs, then load data into Supabase
- Remove ~4.8M lines of static CSV/JSON data files now served from the database
- Extract shared taxa config (`taxa.ts`) and utilities, with `sync.ts` as a thin orchestrator over the pipeline stages
- Use upsert for stable row IDs, with JSONL diff logging for auditing changes
- Add retry logic to GBIF species validation to prevent silent data loss
- Batch Supabase inserts at 50K rows for fast load times

### Server-side API + client-side filtering
- Add 6 API routes backed by Supabase queries (taxa, stats, assessments, species list, species detail, country species)
- Add `redlist_species_query` PL/pgSQL function — single RPC handles pagination, sorting, and priority scoring in one DB round-trip
- Add `POST /api/redlist/species` endpoint and `useRedListSpeciesQuery` React hook with AbortController for instant filter responses
- Implement cross-filtering client-side for responsiveness — species data is fetched per-taxon and gzipped payloads are small (10-20 KB)
- Add partial/composite indexes and 5s statement timeout for query performance
- Wrap `taxa_summary` in RPC for proper access control

### UX improvements
- Defer species fetch until a taxon is selected, making the taxa summary the landing page
- Fetch species per-taxon individually and support taxa multi-select
- Add "All Species" view that bypasses the taxa filter
- Show dimmed table with spinner overlay during pagination instead of replacing content with a spinner

### Migration cleanup
- Consolidate migration files into 4 single-responsibility files:
  - `create_species_table` — table definition + RLS
  - `create_species_indexes` — pg_trgm extension + all indexes
  - `create_taxa_summary_view` — materialized view + refresh/get RPCs
  - `create_redlist_species_rpc` — species query function

## Test plan
- [x] Run `npx tsx scripts/sync.ts` end-to-end against a local Supabase instance
- [x] Run `TEST_TABLE1A=1 npx vitest run scripts/__tests__/validate-table1a.test.ts` to validate taxa_summary against IUCN Table 1a
- [ ] Load dashboard at `/` — verify taxa summary loads as landing page with no cold-start delay
- [ ] Select a taxon group — verify species table loads for that taxon
- [ ] Navigate between taxon groups — verify species table and charts update
- [ ] Apply category, year, country, and observation range filters — verify cross-filtering (each chart excludes its own filter dimension)
- [ ] Test search with partial species names (trigram search)
- [ ] Verify pagination controls and page size changes
- [ ] Test sorting by priority, category, year, and new GBIF observations
- [ ] Verify URL params persist on reload
- [ ] Check NE (Not Evaluated) species toggle works correctly
- [ ] Test "All Species" view shows correct results

## Shane's TODOs:
- [x] Remove duplicate species total row 
- [x] Check naming of 'dashboard' api → renamed to `/api/redlist/species` + `redlist_species_query`
- [x] Check why NE species now includes cows, humans etc. Checked – this isn't new.
- [x] Consolidate the migration files
- [x] Check supabase skill best practices have been applied
- [x] Remove unused api routes
- [x] Save a db dump before any supabase load? Decided this is overkill for now. But was good to experiment with pg_dump and pg_restore.
- [x] Check if we should upsert rather than truncate and insert
- [x] Add unit tests for the upsert logic
- [x] Archive not delete stale rows – decided against this. if species lumps / splits happen, we dont need to lose user data as we can just set ON DELETE SET NULL.
- [x] Think through how to handle species lumping and splitting
- [x] Consider still doing the cross-filters client-side... gzip sends the payloads really small, 10-20mb...
- [x] Switch to diff-based supabase upserts rather than full upserts (feels safer and more controlled). Good chat with Claude. Seems the wrong approach for now. Full upserts are idempotent, so if something goes wrong mid-sync it's okay. Ah I dunno.
- [ ] Check which of the current indexes are unnecessary
- [ ] Check again supabase best practices are applied

Is this overcomplicating things?? Should we just stick to static files? And only add the database later when we need add user interaction?


shaneweisz (commented):
nvm

shaneweisz:
This PR has become too complex. Lots of learnings though:
Key Learnings from the Database PR
The core mistake: premature optimization. The whole Supabase migration was solving a problem that doesn't exist yet. Claude Code's audit confirmed it bluntly — Supabase was being used as a "dumb store" that just replicated fs.readFile(), adding network latency, env var management, a 1,641-line sync pipeline, and an external dependency with no runtime benefit.
Client-side vs server-side is a real tradeoff, but the answer is context-dependent. The 10–20MB payload felt like a problem worth solving, but the actual fix isn't a database — it's progressive disclosure. Show just the taxa summary table on load, then fetch per-taxon data on demand. That solves the transfer size issue without any DB overhead.
Database design requires deliberate upfront thought in a way frontend doesn't. With frontend, validation is cheap — you can see if it works. With schemas and pipelines, bad design decisions compound. AI is useful for frontend; for backend, you still need to own the design yourself.
Static files + Git LFS is the right answer for now. The data is essentially read-only. Linking future user data (assessments, assessors, reviewers) to a scientific name — with synonym resolution in code — is perfectly feasible without a DB. When PostGIS or complex relational data genuinely becomes necessary, migrate then.
Keep PRs small. As the codebase got messier mid-PR, the AI started getting confused too. The PR grew because the scope crept. A cleaner discipline of small, focused PRs would have caught this earlier.
Define acceptance criteria before handing off to the agent. The back-and-forth confusion was partly because the goal kept shifting. Clear upfront criteria — and good tests to verify nothing broke elsewhere — would have prevented a lot of the churn.
The week wasn't wasted. The sync process is better understood, fuzzy species name matching is improved, and the actual constraints (transfer size, progressive disclosure, future PostGIS needs) are now much clearer. That's real progress even if the Supabase branch gets closed.

PR #143, opened 2026-03-15
Add country, family, and order data for NE species via GBIF API

shaneweisz:
### GBIF country data for NE species
New `fetch-gbif-country-data` script populates semicolon-separated ISO country codes for all species in the GBIF CSVs. Always re-fetches all species (no incremental mode — fast enough at ~6 min).

### Kingdom-level query optimization
Instead of querying per-species (~660k API calls, ~45 hours), queries at the kingdom level (Animalia/Plantae/Fungi/Chromista) per country and filters against known species. Reduces to ~1,000 API calls (~6 minutes for a full sync). Single-taxon runs only query the relevant kingdom(s).

### Fix stale fish order keys
Removed 19 non-fish keys (Foraminiferida, Tricladida, nematodes, etc.) and added 26 missing fish orders (Acipenseriformes, Lamniformes, Rajiformes, etc.). Added Myxini (hagfish) class key and Petromyzontiformes (lampreys) order key. ~1,500 fish species newly matched as a result.

### Family & order for NE species
Extended the GBIF CSV with `class_name`, `order_name`, and `family` columns, extracted from the `/v1/species/{key}` response already fetched during validation (zero extra API calls).

### Basis of record filter changes
- **Excluded** `PRESERVED_SPECIMEN` (historical ranges from centuries ago) and `MATERIAL_CITATION` (literature references that may include fossil taxa) from the sync pipeline, API routes, and GBIF link filters
- They remain visible in the occurrence map but default to unselected
- Investigated including them but found they pull in fossil species (e.g. eurypterids in Merostomata) and centuries-old collection records

### Sync pipeline improvements
- Reordered: country data (Phase 4) before new counts (Phase 5)
- `fetch-gbif-new-counts` stays on year-based bucketing (~2,730 calls). Investigated `modified` filter but GBIF's `dc:modified` is the publisher's last-edit date, not when the record appeared in GBIF
- Added `kingdomKey` to taxon config so country fetching only queries relevant kingdoms for single-taxon runs
- CSV column order updated for readability

### NE species loading fix
Removed prefetch of all ~550k NE species on mount which caused Vercel OOM errors. Species are now fetched per-taxon only. Also fixed React setState-during-render warning in NewAssessmentsView.

## Test plan

- [x] Run `npx tsx scripts/fetch-gbif-country-data.ts horseshoe_crabs` — 4 species, correct countries, no fossils
- [x] Run `npx tsx scripts/fetch-gbif-country-data.ts mushrooms` — single kingdom (Fungi), 249 API calls, 1m 42s
- [x] Run `npx tsx scripts/sync.ts horseshoe_crabs brown_algae` — full pipeline, 4m 54s
- [x] Run full sync all taxa — 34m 53s, all phases successful
- [x] Spot-checked 5 new fish mappings (sturgeons, toadfish, etc.) — all correct
- [x] Verified NO_GBIF_DATA species are genuinely missing qualifying GBIF data (no coverage gaps)
- [x] Verify preview deployment no longer OOMs
- [x] Verify NE species load per-taxon in the UI
- [x] Verify occurrence map shows material citation / preserved specimen toggles (defaulted off)


Marothi-Mohale:
This was a very good read. The kingdom-level query strategy is a clever optimization, but what impressed me most was how much care went into the data quality side too , especially the stale fish order fixes and the reasoning behind excluding historical/fossil-skewing record types from the main sync flow. Feels like one of those PRs where both the system design and the domain thinking are doing real work.

Issue #58, opened 2026-02-20
Bug: machine observations for Achoerodus viridis (Eastern Blue Groper) not showing on map


shaneweisz:
Screenshot showing issue:

<img width="1324" height="526" alt="Image" src="https://github.com/user-attachments/assets/a77ca9bc-e1a4-43df-9b5c-e1df651fa9a4" />

[Achoerodus viridis](https://www.iucnredlist.org/species/187572/8572139)
Eastern Blue Groper

Whereas GBIF shows they do have coordinates:

<img width="1197" height="654" alt="Image" src="https://github.com/user-attachments/assets/a4e4a9b7-c288-4f8b-97ab-4145b0201b8a" />

PR #214, opened 2026-04-01
Migrate occurrence map from react-leaflet to MapLibre GL JS

shaneweisz:
## Summary

- **Replace Leaflet's DOM/SVG rendering with MapLibre's WebGL pipeline** to support upcoming IUCN range maps, AOH GeoJSONs, habitat maps, and elevation/DEM layers
- **5000 DOM elements → 1 GPU draw call**: Individual `<CircleMarker>` React components replaced with a single GeoJSON `Source` + `circle` `Layer` with data-driven paint expressions for color, radius, stroke, and opacity
- **50ms polling → event-driven sync**: Split-view map synchronization via shared `viewState` instead of the `MapSync` polling hack
- **Delete `FitBounds.tsx`** — replaced by direct `map.fitBounds()` via ref with `onLoad` callback to handle the race between data fetch and map mount
- **Remove unused cog popover features** — shape-by-type markers, spatial deduplication, uncertainty radius circles, and the entire "More" settings popover (-346 lines). Date-based coloring is now always on, with a toggle to switch to before/after assessment year coloring
- **Fix tooltip clipping** — tooltips now portal to `document.body` with fixed positioning instead of the map container, avoiding `overflow:hidden` clipping
- **Fix InatContributorsChart rerender** — memoized with `useMemo` to prevent re-rendering on every hover state change
- **WorldMap & TradeFlowMap unchanged** — react-simple-maps stays for SVG-based choropleth/arc views

### Files changed

| File | Change |
|------|--------|
| `OccurrenceMapRow.tsx` | Core rewrite: react-leaflet → react-map-gl/maplibre, remove unused cog features, add before/after color toggle |
| `LocateControl.tsx` | `useMap()` from react-leaflet → react-map-gl/maplibre |
| `MapOccurrenceTooltip.tsx` | `latLngToContainerPoint()` → `map.project()`, portal to body with fixed positioning |
| `MapImageTooltip.tsx` | Same portal fix, adapted for MapLibre |
| `FitBounds.tsx` | **Deleted** |
| `layout.tsx` | CSS import swap: leaflet → maplibre-gl |
| `package.json` | +maplibre-gl, +react-map-gl, -leaflet, -react-leaflet, -@types/leaflet |

### Why MapLibre

| Concern | Leaflet | MapLibre GL JS |
|---------|---------|----------------|
| Large polygon rendering (range maps, AOH) | SVG/DOM — sluggish | WebGL — handles millions of vertices |
| Raster/DEM layers (elevation, habitat) | Basic raster only | Native terrain/DEM, hillshade layer |
| Vector tiles | Requires abandoned plugins | First-class MVT support |
| Layer compositing | Limited, CSS-based | Style spec with expressions, data-driven styling |
| Multiple overlay layers | Each adds DOM overhead | All GPU-composited in one canvas |

## Test plan

- [x] Verify occurrence map renders with all 3 basemaps (Streets, Satellite, Terrain)
- [x] Verify map auto-zooms to occurrence bbox on load (e.g. Panthera leo centers on Africa)
- [x] Verify circle markers show correct date-based color gradient
- [x] Test before/after assessment date color toggle in legend
- [x] Test hover tooltips are not clipped at the top of the map
- [x] Test click-to-open-GBIF on markers
- [x] Test linked brushing: hovering a year in histogram dims non-matching markers
- [x] Test linked brushing: hovering a record type in filter dims non-matching markers
- [x] Test split view: synchronized pan/zoom between before/after panels
- [x] Test split view: fitBounds works correctly on initial load
- [x] Test split view: iNat photo hover only shows on the correct panel
- [x] Test temporal animation playback (play/pause/scrub)
- [x] Test all filter controls (basis of record, uncertainty, year range)
- [x] Test LocateControl (geolocation button)
- [x] Verify iNat photo hover highlights marker on map
- [x] Verify InatContributorsChart doesn't flicker during linked brushing
- [x] Verify WorldMap and TradeFlowMap still render correctly (unchanged)
- [x] Run `npm run test` — all 340 tests pass
- [x] Run `npx tsc --noEmit` — no type errors

https://claude.ai/code/session_01P72WmaxNKtGXDdqckp7LAM


app/src/components/OccurrenceMapRow.tsx
```diff
@@ -1277,126 +1427,75 @@ export default function OccurrenceMapRow({
               </div>
             </div>
           ) : mounted ? (
-            <MapContainer
-              center={[20, 0]}
-              zoom={2}
-              style={{ height: "100%", width: "100%" }}
+            <MapGL
+              ref={panelId === "main" || !splitView ? mapRef : undefined}
```
shaneweisz: Bug: In split view, `mapRef` is never assigned to either panel.

When `splitView` is `true`, `renderMapPanel` is called with `panelId="before"` and `panelId="after"`. The condition here is `panelId === "main" || !splitView` — both sides are false, so `ref` is `undefined` for both panels. This means:

1. `fitMapToBbox` (which reads `mapRef.current`) will never work in split view — the map won't auto-zoom to the occurrence bbox.
2. `handleMapLoad` is also gated by the same condition, so pending bbox fitting won't fire either.

Suggested fix: assign the ref to at least one of the split panels (e.g. the "before" panel):
```tsx
ref={panelId === "main" || panelId === "before" ? mapRef : undefined}
```
And similarly for `onLoad`:
```tsx
onLoad={panelId === "main" || panelId === "before" ? handleMapLoad : undefined}
```

---
_Generated by [Claude Code](https://claude.ai/code)_

app/src/components/OccurrenceMapRow.tsx
```diff
@@ -1277,126 +1427,75 @@ export default function OccurrenceMapRow({
...
+              {...mapProps}
+              style={{ width: "100%", height: "100%" }}
+              mapStyle={BASEMAP_STYLES[basemap].style}
+              interactiveLayerIds={[`occ-circles-${panelId}`]}
```
shaneweisz, line 1305: Bug: When `shapeByType` is enabled, the `Source`/`Layer` with id `occ-circles-{panelId}` is not rendered (line 1450: `{!shapeByType && ...}`), but `interactiveLayerIds` still references it. MapLibre will log a warning about a missing layer, and more importantly, click/hover handlers (`handleMapClick`, `handleMapMouseMove`) that rely on `e.features` being populated via `interactiveLayerIds` will stop working — you'll get no features on click/hover.

Fix: conditionally set `interactiveLayerIds` based on `shapeByType`:
```tsx
interactiveLayerIds={shapeByType ? [] : [`occ-circles-${panelId}`]}
```

---
_Generated by [Claude Code](https://claude.ai/code)_

app/src/components/OccurrenceMapRow.tsx
```diff
@@ -1263,6 +1364,55 @@ export default function OccurrenceMapRow({
...
+      id: `occ-uncertainty-${panelId}`,
+      type: "circle" as const,
+      paint: {
+        "circle-radius": [
+          "interpolate", ["exponential", 2], ["zoom"],
+          0, 0,
+          20, [
+            "/",
+            ["*", ["get", "uncertaintyMeters"], 1],
+            0.075, // approximate meters-per-pixel at zoom 20
```
shaneweisz: Bug: The uncertainty circle radius formula is incorrect. The old Leaflet code used `Circle` with `radius` in **meters**, which Leaflet projected correctly at all zoom levels. This MapLibre replacement attempts to approximate it with a pixel-based `circle-radius`, but the formula is wrong in two ways:

1. **Latitude dependence**: meters-per-pixel varies by latitude (Mercator projection), but this formula uses a single constant (0.075). At the equator the circles will be roughly correct, but at 60° latitude they'll be ~2x too large.

2. **The constant itself**: at zoom 20, 1 pixel ≈ 0.075m at the equator, so a 1000m uncertainty circle would be `1000 / 0.075 = 13,333 pixels` — that's a circle filling the entire screen at zoom 20, which is clearly wrong. The formula divides meters by meters-per-pixel to get pixels, but the exponential interpolation from zoom 0 to 20 means at intermediate zooms the values won't scale correctly either.

Consider using a `fill` layer with actual GeoJSON circle polygons (generated with e.g. `turf.circle()`) instead, which would correctly project in meters at all latitudes and zoom levels. Alternatively, use the well-known formula:
```
pixelRadius = metersRadius / (cos(lat) * 0.075 * 2^(20 - zoom))
```
expressed as a MapLibre expression. Though this is complex, getting it right matters since the uncertainty radius is a core analytical feature.

---
_Generated by [Claude Code](https://claude.ai/code)_

app/src/components/OccurrenceMapRow.tsx
```diff
@@ -1263,6 +1364,55 @@ export default function OccurrenceMapRow({
   ) => {
     const panelNewRecords = panelOccurrences.filter((o) => isNewRecord(o.properties.eventDate));
     const panelOldRecords = panelOccurrences.filter((o) => !isNewRecord(o.properties.eventDate));
+    const styledGeoJson = buildStyledFeatureCollection(panelOccurrences);
```
shaneweisz, line 1259: Performance: `buildStyledFeatureCollection` is called inside `renderMapPanel`, which runs on every render. This rebuilds the entire GeoJSON FeatureCollection (iterating all occurrences, computing colors, creating new feature objects) on every mouse move, hover state change, or any other state update — even when the inputs haven't changed.

In the old Leaflet code, each `CircleMarker` was an independent React component, so React's reconciliation handled skipping unchanged markers. Here, the entire collection is rebuilt as a single object.

Consider memoizing the result with `useMemo`. For the single-view case you could do:
```tsx
const styledGeoJson = useMemo(
  () => buildStyledFeatureCollection(filteredOccurrences),
  [filteredOccurrences, buildStyledFeatureCollection]
);
```
And pass the pre-computed GeoJSON into `renderMapPanel` instead of computing it inside. For split view, you'd memoize two separate collections.

---
_Generated by [Claude Code](https://claude.ai/code)_

app/src/components/OccurrenceMapRow.tsx
```diff
@@ -1233,27 +1177,184 @@ export default function OccurrenceMapRow({
...
+          _strokeColor: strokeColor,
+          _radius: radius,
+          _strokeWidth: strokeWidth,
+          _category: category,
+        },
+        geometry: feature.geometry,
+      };
+    });
+    return { type: "FeatureCollection", features };
+  }, [hoveredObs, hoveredType, hoveredYear, hoveredFeature, colorByDate, minDateNum, maxDateNum, isNewRecord]);
```
shaneweisz: Missing feature: The old Leaflet code had **dimming** logic — when hovering a year or type in the legend, non-matching markers were rendered with `opacity: 0.15`, smaller radius (4px), and thinner stroke. This made the brushed markers visually pop.

In the new code, `isDimmed` / `isTypeDimmed` are gone entirely. Non-brushed markers keep their normal appearance (radius 5, opacity 0.9, stroke 2). This significantly reduces the effectiveness of the brushing interaction — the highlighted markers don't stand out nearly as much when everything else stays at full opacity.

Consider adding dimming back. With the GeoJSON/data-driven approach, you could add `_opacity` to the feature properties and use `["get", "_opacity"]` for `circle-opacity`, similar to how you handle radius and color.

---
_Generated by [Claude Code](https://claude.ai/code)_

app/src/components/OccurrenceMapRow.tsx
```diff
@@ -1277,126 +1427,75 @@ export default function OccurrenceMapRow({
...
-                  } else {
-                    strokeColor = "#6b7280";
-                    fillColor = "#9ca3af";
-                  }
-                } else {
-                  strokeColor = isNew ? "#15803d" : "#6b7280";
-                  fillColor = isNew ? "#22c55e" : "#9ca3af";
-                }
-                const inatMatch = inatPhotosByGbifId.get(feature.properties.gbifID);
+                const styledFeature = styledGeoJson.features[idx];
```
shaneweisz: Performance concern in shape-by-type mode: When `shapeByType` is enabled, you're still computing `styledGeoJson` via `buildStyledFeatureCollection` (line 1367) just to read `_fillColor`/`_strokeColor`/`_category` from it by index. But then you also render N individual `MapLibreMarker` components. For 1000 markers this creates 1000 DOM-backed Marker overlays, which is exactly the DOM-heavy pattern this migration is trying to move away from.

Two suggestions:
1. Short-term: Skip the `buildStyledFeatureCollection` call when `shapeByType` is true and compute colors inline (you already have the logic in `buildStyledFeatureCollection` — just extract it as a helper function).
2. Longer-term: Consider using MapLibre's `symbol` layer with `icon-image` for the shape markers, generating the SVGs as image sprites added to the map style. This would keep everything GPU-rendered.

---
_Generated by [Claude Code](https://claude.ai/code)_

app/src/components/OccurrenceMapRow.tsx
```diff
@@ -840,7 +818,8 @@ export default function OccurrenceMapRow({
   const [shapeByType, setShapeByType] = useState(false);
   const [splitView, setSplitView] = useState(false);
   const [splitDate, setSplitDate] = useState<string>(assessmentDate?.split("T")[0] || "");
-  const mapSyncRef = useRef<{ center: [number, number]; zoom: number } | null>(null);
+  const [sharedViewState, setSharedViewState] = useState({ longitude: 0, latitude: 20, zoom: 1.5 });
+  const mapRef = useRef<MapRef>(null);
```
shaneweisz, line 743: Nit: `sampleSize` default changed from 300 to 1000. This is combined with removing the stratified sampling logic in `route.ts`. With the larger default + no stratified sampling, the initial load will be slower and the API will make more paginated requests to GBIF. If this is intentional, it might be worth noting in the PR description.

---
_Generated by [Claude Code](https://claude.ai/code)_

app/src/components/OccurrenceMapRow.tsx
```diff
@@ -840,7 +818,8 @@ export default function OccurrenceMapRow({
...
   const [dedupGrid, setDedupGrid] = useState(0.01); // ~1km
   const [sampleSize, setSampleSize] = useState(1000);
```
shaneweisz, line 744: Two behavioral changes in this block worth calling out:

1. **`colorByDate` defaults to `true`** (line 815, was `false` as `colorByYear`). This changes the default marker appearance — markers will show a date gradient instead of green/gray "new vs old assessment" coloring.

2. **`sampleSize` defaults to `1000`** (line 824, was `300`). Combined with removing the stratified sampling logic in `route.ts`, the initial load will fetch ~3x more records from GBIF.

Both may be good changes, but worth noting as intentional in the PR description.

---
_Generated by [Claude Code](https://claude.ai/code)_

shaneweisz (commented):
Thorough review of the Leaflet → MapLibre GL JS migration. The overall direction is solid — the move to WebGL rendering with a GeoJSON Source + circle Layer is the right architecture, the split-view sync via shared viewState is much cleaner than the old polling hack, and the fitBounds-via-ref pattern is a good replacement for the FitBounds component.

However, there are several bugs and regressions that should be addressed before merging:

### Bugs (must fix)

1. **Split-view `mapRef` never assigned** — neither the "before" nor "after" panel gets the ref, so fitBounds doesn't work in split view.

2. **`interactiveLayerIds` references non-existent layer in shape-by-type mode** — when `shapeByType` is true, the circle layer isn't rendered but is still listed in `interactiveLayerIds`, breaking click/hover on the map.

3. **Uncertainty circle radius formula is incorrect** — the pixel-based approximation doesn't account for latitude (Mercator distortion) and the scaling math produces wildly wrong values. The old Leaflet `Circle` handled meters-to-projection correctly. Consider using `turf.circle()` to generate actual GeoJSON polygons.

### Missing features (should fix)

4. **Dimming of non-brushed markers removed** — when hovering a year or type in the legend, the old code dimmed non-matching markers to 15% opacity with smaller radius. This is completely gone, making the brushing interaction much less effective.

### Behavioral changes (should document / confirm intentional)

5. **`colorByDate` defaults to `true`** (was `false`) — changes the default appearance for all users.
6. **`sampleSize` defaults to `1000`** (was `300`) — 3x more GBIF records fetched on initial load.
7. **Stratified sampling removed from `route.ts`** — the logic that ensured underrepresented `basisOfRecord` types were included in the sample is gone. This was added to fix a specific issue (GitHub #58) where e.g. MACHINE_OBSERVATION records were absent from samples.

### Performance (nice to have)

8. **`buildStyledFeatureCollection` is called every render** inside `renderMapPanel` — should be memoized.
9. **Shape-by-type still creates N DOM markers** — same DOM-heavy pattern the migration aims to eliminate. Longer-term, consider symbol layers.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #257, opened 2026-06-06
Add /browse + llms.txt so LLMs can answer questions from the dashboard

shaneweisz:
## Why

The dashboard is a client-rendered SPA, so pointing an LLM (Claude/ChatGPT web UI) at the live URL returns an **empty HTML shell** — no data. And threat filtering (e.g. climate change = code `11`) only runs client-side, so there's no single URL today that answers *"which coral species are threatened by climate change?"*

This adds a small **server-rendered web layer** so an LLM can answer base-filter questions from a fetched URL, navigating the data like a human would — without needing to know internal codes on the way in *or* out.

## What

- **`/browse`** — mirrors the dashboard's own filter-param vocabulary; returns readable **HTML** (or **JSON** via `?format=json`).
  - Accepts plain-English values: `threats=climate-change`, `categories=endangered`/`threatened`, `taxa=birds`, `countries=Brazil` (codes still work too).
  - Renders codes back as **labels** (threats + IUCN categories).
  - Leads with a **total + by-category breakdown**; **capped at 200 rows** (browsable, not a bulk dump).
  - Self-describing **vocabulary page** when called bare; **search-only** fast path via the search index.
- **`/llms.txt`** — concise, generated guide to the vocabulary + example URLs (the thing you paste into a chat).
- **Shared `species-filter` predicate** extracted from `RedListView` so the SPA and `/browse` stay in sync (OR-within / AND-across / prefix threats — behaviour unchanged).
- **`filter-vocab`** — human-friendly resolvers + `threatDisplay` (deep sub-codes like `5.4.1` walk up to the nearest known label).
- **`countries`** — ISO name↔code maps relocated out of the client `WorldMap` so server code can import them; `WorldMap` re-exports `ALPHA2_TO_NAME` so existing importers are untouched.
- **`layout`** — small server-rendered footer linking `/browse` + `/llms.txt` (the only crawler-visible spot, since the app is client-rendered).

Discoverability preserved (no `noindex`); no bulk-downloadable data file.

## Example

`GET /browse?taxa=corals&threats=climate-change` → **878** corals, breakdown `CR:31 EN:244 VU:56 NT:18 LC:417 DD:112`, top hit *Acropora cervicornis (Critically Endangered)* with labelled threats.

## Verification

- Live-tested against real R2 data: canonical corals+climate query, `Brazil`→`BR` resolution, cross-taxa `search=tiger`, empty/unknown/over-cap/dedup cases.
- **27 new tests**; full suite **438/438** passes; typecheck + lint clean.

## Notes / deliberate limits

- **Cross-taxa listings** (all groups at once) intentionally aren't a 280k scan — the index/`llms.txt` steer per-taxon. Could be added via the precomputed summaries if wanted.
- **Synonym search** was dropped: `SpeciesRow` has no `synonyms` field, so search matches scientific + common name (same as the dashboard). Adding it needs a data-model change.

### DIFF
--- a/app/src/app/browse/__tests__/browse.test.ts
+++ b/app/src/app/browse/__tests__/browse.test.ts
@@ -2,9 +2,15 @@ import { describe, it, expect, vi, beforeEach } from "vitest";
 import { NextRequest } from "next/server";
 
 // Mock the data store so the route logic can be tested without CSV data.
+// isOutdated keeps its real >10-year rule (it's a pure function).
 vi.mock("@/lib/data/species-store", () => ({
   getSpecies: vi.fn(() => []),
   searchSpecies: vi.fn(() => []),
+  isOutdated: (date: string | null) => {
+    if (!date) return true;
+    const year = parseInt(date.slice(0, 4), 10);
+    return Number.isNaN(year) ? true : new Date().getFullYear() - year > 10;
+  },
 }));
 
 import { getSpecies, searchSpecies } from "@/lib/data/species-store";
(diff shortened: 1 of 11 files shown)


shaneweisz:
Superseded by #291. That PR was built on the pre-Phase-2 in-memory species path (`getSpecies`/`SpeciesRow`), removed in #269, and was 124 commits stale. #291 re-implements /browse + llms.txt cleanly on the live DuckDB read layer (querySpecies/searchSpecies), salvaging the data-layer-independent pieces (filter-vocab, countries, species-filter + tests), and adds arbitrary-rank browse + synonym-aware lookup.

PR #280, opened 2026-06-12
universe-names search index — fast cold CoL search

shaneweisz:
Follow-up to #279 (CoL universe search): that fallback scanned `species/` (15 partition files, all columns) from R2 — ~1.4s warm but **~12s on a cold container**. This adds a compact, name-sorted index so the CoL fallback is fast even cold.

## What
- **`build-universe-names`** (wired into the sync after build-matching) writes `universe-names.parquet`: ONE file, the extant universe (`in_base`, extant, ex-Homo), columns `col_id, scientific_name, taxon_group, name_lower`, **sorted by `name_lower`**, 50k-row groups.
- **`searchSpecies`** CoL fallback now reads it: a **prefix-range** query on the sorted `name_lower` (`>= q AND < q || chr(0x10FFFF)`) — Parquet min/max stats prune to ~1 row group, so it reads a tiny slice even cold. A **substring backstop** runs only when the prefix is sparse (mid-name queries). Graceful no-op if the index isn't in the sync prefix.
- `next.config`: index is R2-only → added to `COL_ARTIFACTS`.

## Validated (local)
Prefix (`Aeschnosoma yelenae`), genus (`Aeschnosoma`), mid-name substring (`yelenae` → matches), and assessed (`Panthera leo`) all correct; NE → leaf node. 361 tests.

## Measured cold win
Before/after on the deploy in the PR comments.


shaneweisz:
## Measured (preview deploy, R2)

| | #279 (scan species/) | #280 (sorted index) |
|---|---|---|
| **warm** (steady state) | ~1.4 s | **~1.3 s** |
| **cold** (first search, fresh container) | ~13.9 s | **~8.4 s** |

- **Pruning confirmed** over R2 httpfs (isolated test): the prefix-range query reads **1 of 40 row groups** in **1.2 s cold / 0.84 s warm**; row-group `name_lower` min/max are clean and ordered.
- The residual cold (~8.4 s) is **serverless cold-start + Vercel's first-R2-read latency**, not data volume — the index read itself is ~1.2 s. `/api/search/warm` (called on page load) primes the index footer to absorb some of it; after warm, fresh CoL searches are ~1.3 s.
- Substring (mid-name) backstop works and runs only when the prefix is sparse.

Net: warm unchanged (fast) with confirmed pruning so it scales; cold ~40% better; 1 file vs 15, simpler. Getting cold to sub-second would need a keep-warm cron (separate) — the index can't beat the serverless/R2 cold floor on its own.

shaneweisz:
Closing — not worth the added complexity. The win is cold-only (~13.9s → ~8.4s) and doesn't cross a meaningful threshold (8.4s still feels broken), while warm is unchanged (~1.3s) and "pruning scales" is moot for a fixed ~2M dataset. The NE→leaf-node correctness fix is already in #279 (merged); this PR was purely the perf index. Net: a new build step + sync phase + a 50MB R2 artifact duplicating names already in species/, for no threshold-crossing benefit. If the occasional cold matters later, a keep-warm cron is the cleaner lever (makes cold actually fast, no new artifact, also helps the NE-list cold-start).

PR #359, opened 2026-07-07
Add SSC groups mode: pilot view for IUCN mammal Specialist Groups

shaneweisz:
## Summary
- Adds an "SSC groups mode" toggle next to "Table 1a mode" — a flat summary table organized by IUCN SSC Specialist Group (Cat SG, Cetacean SG, Small Mammal SG, etc.) instead of taxonomic order, so a specialist group can pull up exactly the species it's responsible for. Clicking a group drills into its filtered species list, same as any other sub-group.
- **Pilot scope**: the 35 mammal-taxon SSC groups, plus a 36th **"No SSC Group" remainder row** — mammal species not claimed by any of the 35 (via an exclude-based filter covering all 35 groups' scopes). Also clickable. Its numbers sum with the other 35 to exactly the real Mammals total, confirming no gaps or double-counts — and it surfaces genuinely useful things: the wild Bactrian camel (deliberately excluded from Wild Camelid SG's South-American-only remit), plus real SSC coverage gaps (pygmy right whale/Neobalaenidae, pronghorn/Antilocapridae, musk deer/Moschidae, red panda/Ailuridae).
- Each group's species filter was sourced from its own IUCN page (https://iucn.org/our-union/commissions/group/1445 directory), not just inferred from its name — including tricky boundary splits like Bear SG vs. Polar Bear SG, Antelope SG vs. Caprinae/Bison/Wild Cattle SG (all carved out of Bovidae), and Otter SG vs. Small Carnivore SG (carved out of Mustelidae).
- Extends `SpeciesFilter` with genus/species-level matching (`genera`, `excludeGenera`, `speciesNames`, `excludeSpeciesNames`) to express those splits, threaded through `matchesFilter`, `matchesTaxonomyFilter`, and `filterToSql` so client-side filtering, the assessor/reviewer candidate routes, and the summary-build pipeline all agree.
- The 36 groups live under a new `ssc-groups` tree node kept separate from the real `mammals` node, so they don't clutter the normal Mammals drill-down.
- Rows are sorted by # Red List Assessed descending (busiest groups first, remainder pinned to the bottom).
- All three mode toggles (Table 1a mode, SSC groups mode, # Described source) are now reachable on mobile, not just desktop.
- Table 1a mode / SSC groups mode is now URL-synced (`?layout=table1a|ssc`) — survives reload/share, the browser back button correctly returns to the flat table after drilling into a group, and clicking the "Mammals" breadcrumb row after drilling out of SSC mode returns to it instead of falling through to the plain taxon tree.
- Each section's subtotal is now merged into its header row (title + numbers together) instead of a separate row after all the group rows — visible immediately without scrolling.

## "# Described Species" now defaults to Catalogue of Life almost everywhere
Only the 8 top-level summary taxa and Table 1a's 21 rows have a genuinely IUCN-sourced described-species figure (the Table 1a PDF). Every other node — Rodents, Bats, Beetles, all 36 SSC groups, every reptile/amphibian/fish/insect sub-split — was showing a static third-party citation (MDD, Reptile Database, AmphibiaWeb, Eschmeyer, Zhang 2011) or a hand-typed approximation, none of which get re-verified as species data changes. Switched every non-official node to default to `colDescribed`, which is recomputed from the live CoL backbone on every sync — real automated maintenance instead of numbers that quietly go stale. The manual IUCN/CoL toggle still works exactly as before for the 8 summary taxa + Table 1a rows.

This surfaced real CoL data-quality issues along the way, all investigated and either fixed or documented:
- **Bison SG matched 0 CoL species** — CoL lumps genus *Bison* entirely into *Bos* (verified directly against the backbone). Added a species-name override, and excluded those same two names from Wild Cattle SG's computation so it doesn't double-count.
- That surfaced **domestic breeds inflating counts** more broadly (*Bos taurus*, *Canis familiaris*, *Equus caballus/asinus*, *Felis catus*, *Vicugna pacos*, *Bos frontalis*) — excluded all 7, each verified to have its wild-form sibling species present separately in CoL so excluding the domestic form doesn't lose real coverage.
- **Apparent CoL coverage gaps** in a handful of groups genuinely undercounted vs. what IUCN has assessed. Originally papered over with a fallback to the old static estimate whenever `colDescribed < totalAssessed` — removed that fallback (see below): showing the true, if surprising, CoL number is more honest than silently reverting to a number that never gets re-verified either. (Turned out most of the actual "gaps" were a stale local dataset, not real CoL gaps — see the CoL backbone refresh section below.)

## Making "# Described Species" verifiable
A skeptical SSC group member has no way to check a CoL-derived number from the UI alone, so the "# Described Species" info icon is now click-to-open (not hover — see below) and explains, in plain language, exactly what's being counted:
- The node's `SpeciesFilter` rendered as readable text, e.g. *"Family: Otariidae (14), Phocidae (18), Odobenidae (1)"* — every single taxon name listed in full, not truncated, each **linked straight through to a Catalogue of Life search scoped to that taxon** (extant + accepted/provisionally-accepted species only — the same definition `colDescribed` itself counts), with **each name's own share of the total shown in brackets** so the breakdown is checkable at a glance, not just the aggregate.
- A note explaining any CoL-specific quirk for that node (e.g. Bison SG: "Catalogue of Life classifies Bison within genus Bos, so this count matches by species name instead of genus").
- A link to Catalogue of Life as the source, plus a pointer to click the row itself to see the underlying species list (and toggle "Unassessed" to see the rest) — reusing the existing, well-tested drill-down rather than building new auto-navigation into the heavier "Unassessed" view mode.

IUCN-cited rows keep showing their original citation (e.g. "IUCN 2025-2" or "Mammal Diversity Database") via the same info icon, so every row now explains its number, whichever source it came from.

**Data plumbing for the taxon-name links:** `scripts/build-col-taxon-ids.ts` mines every taxon name referenced anywhere in the taxonomy tree's filters (177 names), and resolves each to its Catalogue of Life taxon id against `backbone.parquet` — which already carries an id for every rank (order/family/genus/species), not just species, so this is a lookup against data we already sync locally, no new fetch. 167/177 (94%) resolve cleanly; the rest (e.g. Bison, lumped into Bos in this CoL release) are left as plain, unlinked text rather than link to a guess. Output (`src/config/col-taxon-ids.json`, ~180 entries) is small and derived from committed source, so — unlike the other sync outputs — it's committed to git, not published to R2; wired into `scripts/sync.ts` as a new phase, and also re-runnable standalone whenever a node's filter changes.

**Link target — search, not the static taxon page:** a container-rank name (class/order/family/genus) links to `catalogueoflife.org/data/search?TAXON_ID=<id>&extinct=0&rank=species&status=accepted&status=provisionally%20accepted`, not the bare taxon profile page, so clicking through shows exactly the species list being counted. Deliberately minimal query string — dropped the `facet=…` params a copy-pasted CoL search URL carries (`facet=rank`, `facet=status`, `facet=nomStatus`, …, `facetLimit=50`, `limit=50`, `offset=0`): these read as sidebar-filter-widget configuration (which facets the UI pre-expands), not applied result filters, so removing them should be inert. **Caveat**: I couldn't verify this empirically — automating a real browser against catalogueoflife.org hit a 403 (their bot detection), so this is reasoned from the URL's parameter structure, not an A/B test. Worth a manual click-through spot-check. Species-rank names (e.g. a single-species SG like Polar Bear SG) link to their own taxon page instead — a species has no species-level children to search for.

**Per-name breakdown:** each linked name's own count (e.g. Otariidae's 14 vs. Phocidae's 18) comes from a new `NodeSummary.colBreakdown` field, computed in `attachColCounts` by re-running the node's filter narrowed to one name at a time — same excludes, same overrides — so the entries always sum to exactly `colDescribed`. Only computed for the 36 nodes whose primary filter dimension lists more than one name (200 extra per-name queries across the tree, only on a full sync); single-name lists skip it since a bracketed count next to the same number already shown as the row total would be redundant.

The plain-language description, the linkification, the per-name breakdown, and the CoL SQL computation are all generated from the same shared config/build step, so they can't silently drift apart the way independently-maintained lists would.

## UX fix: hover tooltip → click-to-open popover
The original hover-only tooltip vanished the instant the cursor left the tiny info icon — there was a gap between the icon and the tooltip box, so moving the mouse toward the tooltip (e.g. to click a link inside it) crossed empty space first, and CSS `:hover` has no grace period. Converted to a click-to-open, portal-rendered popover (mirrors the existing column-visibility menu pattern) that stays open regardless of mouse position, closes on outside click/Escape/scroll, and makes every source citation a real clickable link.

## CoL backbone was 7 months stale — fixed the bug, refreshed the data, refined the extinct-species rule
Cross-checking the new tooltip against the live catalogueoflife.org site (a user spot-check, not something I'd have caught otherwise) surfaced real discrepancies: our Rodentia count was 2,337 vs. the live site's 2,364; Wild Pig SG was 16 vs. 18; Hippo SG's pygmy hippo was missing entirely, which I'd documented as a genuine CoL coverage gap.

**Root cause**: `fetch-coldp.ts` had a hardcoded CoL XR dataset key (`313100` = `COL25.11 XR`, November 2025) instead of resolving the current release. Every future sync would have kept re-fetching that exact same 7-month-old snapshot forever. Fixed:
- `fetch-coldp.ts` now queries ChecklistBank for the newest `origin=xrelease` dataset at fetch time, instead of a hardcoded default.
- `build-backbone.ts`'s `COL_BASE_DATASET` had the same problem (numeric key, one release behind) — switched to CoL's `3LR` rolling-latest alias.
- Refreshed the full local backbone to the current release (`COL26.6 XR`, June 2026) and re-verified: Rodentia and Wild Pig now match the live site exactly, and the pygmy hippo (`Choeropsis liberiensis`) is no longer a synonym in this release — my "missing entirely" claim was wrong, just stale data. Deleted that tooltip note.
- The refresh also split three more domestic forms from their wild siblings (`Bubalus bubalis`/`Bubalus arnee`, `Lama glama`/`Lama guanicoe`, `Sus domesticus`/`Sus scrofa`) — added all three to the domestic-exclude list, same pattern as the existing ones. Wild Camelid SG now reads an exact 2/2.

**Separately, refined what counts as "described" for extinct species.** IUCN's own Red List Guidelines (fetched and read directly — `nc.iucnredlist.org/redlist/content/attachment_files/RedListGuidelines.pdf`) state taxa extinct before 1500 CE are out of scope for assessment entirely, and domesticated taxa are excluded (independently validating the domestic-exclude work above). CoL only gives us a boolean extinct flag, no date, so the closest achievable rule is: **a species counts as described if it's extant, or CoL flags it extinct but IUCN's own linked assessment agrees (category EX or EW)**.

Deliberately *not* "any IUCN assessment exists for this col_id" — `species_link` matches by name/synonym, not taxonomic concept, so a CoL-extinct-flagged name can link to a stale assessment of an unrelated or since-split taxon. Verification surfaced exactly this: several CoL-extinct-flagged names were linked to non-EX assessments, including `Columba arquatrix` (a common, definitely-not-extinct LC pigeon) — a real matching artifact that requiring the category itself be EX/EW correctly excludes. The properly-corroborated set (both CoL and IUCN agree) is 222 species — the Great Auk, the Thylacine, five Mauritian giant tortoise species, the Christmas Island Pipistrelle, and similar well-known recent extinctions. Applied consistently across `build-taxa-summary.ts`'s two aggregate sites and the runtime "Unassessed" species-list query in `species-duckdb.ts`, so the summary numbers and the actual species list never disagree — verified in the browser that Hippo SG's "Unassessed" toggle correctly shows 0 species (not a stray never-assessed fossil). **Hippo SG now reads an exact 4/4.**

This touched 33 data files (the whole CoL backbone + both summary JSONs) — republished to R2, current pinned sync `2026-07-09T13-33-23Z`.

## Fixes along the way (caught by actually driving the feature in a browser)
- **URL persistence**: an SSC selection wasn't surviving reload/share, since `ssc-groups` sits outside the default view's ancestor chain — fixed + covered by a regression test.
- **Breadcrumb rows**: drilling into a group (e.g. Bat Specialist Group) showed bogus "All Species" / "SSC Specialist Groups" rows above it, from the same ancestor-chain issue in a different code path — fixed.
- **Deployment data**: the `prebuild` step overwrites local data with whatever's pinned in `app/latest-sync.txt`, so the precomputed stats file needed regenerating from real data and publishing to R2 or the SSC endpoint would deploy empty. Republished several times more as the pilot grew — current pinned sync is `2026-07-09T13-33-23Z` (see the CoL backbone refresh section above for the most recent one).
- **Mobile toggles hidden**: the Table 1a mode / SSC groups mode / # Described source controls were `hidden sm:flex` — unreachable below the sm breakpoint. Now always visible, wrapping onto multiple lines as needed.
- **Desktop horizontal scroll**: capped the name column's width with ellipsis truncation + a hover tooltip for the full name, and narrowed "# Described Species" (by letting its header wrap instead of forcing one long line) to free up room for the other columns.
- **Group row numbers reading 0**: drilling into any SSC group showed 0 assessed / 0.0% in its own summary row — a regression from the breadcrumb-rows fix above, which had also (accidentally) stopped the background fetch that row's real numbers came from. Fixed by separating "don't render fake breadcrumb rows" from "don't fetch the data" — only the former was ever the actual bug.
- **Back button didn't return to the mode**: entering Table 1a/SSC groups mode never touched the URL, so there was no history state to go back to. Fixed by URL-syncing the mode and atomizing the taxon+subgroup+mode navigation into a single history push instead of two independent ones.
- **Tooltips clipped**: an `overflow-hidden` added to fix the horizontal-scroll issue above also clipped hover tooltips at the time. Reverted to a plain `max-width` cap — imprecise by ~11px due to a `position: sticky` + `table-layout: auto` browser quirk, but that's a fine tradeoff against breaking every tooltip. (Later superseded for the described-species tooltip specifically by the portal-rendered popover above, which isn't clipped by scroll/sticky ancestors at all.)
- **Hover tooltip unclickable**: see "UX fix" above.

## Count vs. displayed species list had drifted apart (another user-found bug)
Testing Cat Specialist Group directly: the "# Described Species" count correctly excluded `Felis catus` (the domestic cat), but clicking through to the actual "Unassessed" species list still showed it — 6 species in the count, 7 in the list. Root cause: the count and the list are computed by two completely separate code paths that had quietly drifted apart.
- **Count**: `build-taxa-summary.ts`'s `filterToSql` (server-side SQL, build time) — applies the domestic-exclude list and the Bison genus-lumping override.
- **Displayed list**: fetched broadly (e.g. all mammals), then narrowed **client-side** via `matchesFilter`/`speciesMatchesNode` (`taxonomy-utils.ts`) — which never knew about either special case.

Fixed by mirroring both into the client-side matcher. The Bison override is an *extra* match path, not a replacement — a test caught that IUCN-sourced rows still say "Bison bison" while CoL-sourced rows say "Bos bison" (CoL's lumped name), so both need to keep matching. Verified in the browser: Cat SG's count and its displayed NE list now both agree on 6 species.

**Separately investigated (confirmed not a bug):** is `Leopardus munoai` incorrectly unmatched to IUCN's `Leopardus colocolo` assessment? No — checked directly via `species_link`: IUCN's `Leopardus colocolo` (NT, assessed 2014) correctly resolves to CoL's currently-accepted `Leopardus colocola` (matched as a synonym). `Leopardus munoai` is a *separate* accepted species in CoL's current taxonomy (the old "Pampas cat" complex has been split into several full species: colocola, munoai, garleppi, and others) with zero entries in `species_link` — a genuine, real assessment gap (a taxonomic split IUCN hasn't individually re-evaluated yet), not a matching failure. Exactly the kind of thing this feature is meant to surface for a specialist group, not paper over.

## Full 36-group audit: two more real bugs, everything else confirmed legitimate
Systematically compared, for every SSC group, the IUCN-assessed species matching that group's filter against their own CoL-linked classification — categorizing every mismatch instead of spot-checking. Found and fixed two more real bugs:
- **Cetacean SG undercounted by 25 species.** CoL's current taxonomy renames the beaked-whale family *Ziphiidae* → **Hyperoodontidae** (28 species) and separately carries **Neobalaenidae** (the pygmy right whale, previously landing in "No SSC Group") — neither was in Cetacean SG's family filter. Added both (mirrored into the "No SSC Group" exclude list too). Cetacean SG: 71 → 96 described, now matching its 93 assessed species exactly.
- **Tooltip CoL-search links didn't match what's actually counted.** The click-through link searched `status=accepted&status=provisionally accepted`, but `colDescribed` itself only counts `status=accepted` (deliberately — provisionally-accepted CoL names overshoot IUCN's own totals by ~26%, per the existing `SPECIES_STATUS` comment). Dropped the provisionally-accepted param so click-through always matches the count it's meant to verify.

Everything else checked out as legitimate, not bugs:
- **Extinct/EX-EW rule**: re-verified per-group (my first verification pass had a bug in *its own* script, not the app — it flagged every CoL-extinct-flagged species as excluded without checking each one's own IUCN category). Corrected, it confirms the rule works everywhere: *Hippopotamus lemerlei* (Hippo SG), *Thylacinus cynocephalus* + 12 others (Australasian Marsupial SG), 7+ *Pteropus* bats, 2 *Dusicyon* foxes, *Rucervus schomburgki*, the Steller's sea cow, 8 *Nesophontes*/*Solenodon*/*Geocapromys*, etc. — all correctly counted as described via confirmed EX/EW. Only 4 species across all 36 groups are correctly *excluded* (CoL flags them extinct, but IUCN's own category isn't EX/EW — e.g. *Mystacina robusta* is CR, not EX).
- **Curated-checklist (Base GSD) coverage gaps**: scattered thinly across many unrelated families (rodents, bats, primates, marsupials) rather than concentrated — the signature of genuine CoL-side lag on recently-described/split taxa (e.g. *Myotis oreias*, *Neohylomys vietnamensis*), not a systematic matching bug.
- A handful of mammal families (Tragulidae, Moschidae, Camelidae-minus-*Lama*/*Vicugna*, Ailuridae, Antilocapridae) genuinely have no dedicated SSC group among the 35 pilot groups and correctly land in "No SSC Group."

## Three more UI bugs (user-reported)
- **Dead "click this row" link**: the described-species popover's "Click this row to see the species list" was plain text inside a portal whose container stops click propagation — clicking it did nothing. Removed the sentence entirely rather than wiring it up — the row itself is already visibly clickable.
- **Assessed/Unassessed toggle reset to the homepage**: the toggle buttons hard-set the URL to `/` or `/?view=new-assessments`, wiping any taxa/subgroup selection. Now preserves the full query string across the toggle, only swapping the `view` param. This also surfaced a second layer of the same bug: even with the URL fixed, a *specific SSC group* selection (e.g. Cetacean SG) still fell back to all of Mammals on toggle, because a `viewMode`-switch effect in `RedListView.tsx` unconditionally cleared `selectedSubgroups` — despite new-assessments mode already fetching a selected sub-group directly (`SPECIES_API?taxon=<subgroup>&category=NE`, see the fetch effect). Stopped clearing it; toggling Unassessed while viewing Cetacean SG now stays on Cetacean SG.
- **"# Described Species" source cited a generic, dateless CoL homepage link**: now cites and links to the *exact* XR release the data was built from (e.g. "Catalogue of Life (COL26.6 XR)", linked via its DOI). `fetch-coldp.ts` writes this citation metadata to `src/config/col-release.json` on every sync (checked into git, same pattern as `col-taxon-ids.json`) so it can never silently drift from the data actually in use.

## Expandable Assessed/Not Evaluated breakdown in the popover (requested)
Turned the flat per-name breakdown (`Order: Rodentia (2388), Eulipotyphla (589), Scandentia (23)`) into an expandable list, so an SSC member can see exactly where a group's discrepancy lives without leaving the tooltip:
```
Order:
▾ Rodentia (2388)
    Assessed (2302)
    Not Evaluated (86)
▸ Eulipotyphla (589)
▸ Scandentia (23)
```
Clicking Assessed/Not Evaluated navigates to that node's own species list in the matching view, **narrowed to just the clicked name** (clicking Rodentia's "Not Evaluated" shows only its 86 species, not Small Mammal SG's full 160) — no new API endpoint needed, since species are already fully fetched per node client-side; the narrowing is one more filter stage (`bd=<nodeId>:<rank>:<name>` URL param) in `RedListView`'s existing `taxaFilteredSpecies` pipeline, shown as a removable chip. `build-taxa-summary.ts`'s per-name breakdown query now also computes each name's `colNe` (mirroring the existing node-level query), stored as `NodeSummary.colBreakdown[].neCount`.

Gated so a stale `bd=` can't silently hide a whole group's species after navigating elsewhere: the filter only applies while its own `nodeId` is still in `selectedSubgroups`, and any direct taxa/subgroup change drops it outright. Caught by testing: the existing view-mode-switch effect calls `setSelectedTaxa` as a conditional no-op on every Assessed↔Unassessed toggle (clearing an `"all"` selection if present) — my first pass cleared `bd=` on *every* call to that setter, including this no-op, which clobbered the very filter a popover click had just set in the same navigation. Fixed by only dropping `bd=` on an actual (reference-changed) value.

## Single-name breakdowns too, and flagging CoL/IUCN assessed-count mismatches (requested)
The breakdown above only ever attached to nodes with 2+ names in their primary dimension (e.g. Small Mammal SG's 3 orders) — a node like Primate SG (`orderNames: ["primates"]`, one name) still showed a static "Order: Primates" line with no expand, even though it's the more common shape (most SSC groups have just one family/order). Removed the 2+-name restriction so every node with a resolvable primary dimension gets the same expandable Assessed/Not Evaluated breakdown — this is what revealed Primate SG isn't actually a clean 0-NE group despite reading 100% assessed at the summary level: expanding it shows `Assessed (521)` / `Not Evaluated (6)`, plus its own IUCN-mismatch flag (`trueAssessed` 527 vs. 521) — see below.

That in turn surfaced a second, more useful signal: does the CoL-derived "Assessed" count for a name actually match how many species IUCN itself has assessed under that name? Added `colBreakdown[].trueAssessed` — IUCN's own assessed count for that one name, computed the same way `count`/`neCount` are but against `assessed.parquet`'s own class/order/family fields instead of CoL's. (One subtlety: computing `trueAssessed` deliberately omits `nodeId` from the shared `filterToSql` call — Bison SG's CoL species-name override, which replaces the normal genus match with a CoL-specific lumped-name list, only makes sense for CoL-sourced rows; running it against IUCN's own data would wrongly return 0, since `assessed.parquet` still says "Bison bison", never CoL's "Bos bison".) This first pass surfaced the mismatch with a warning icon — superseded below by a proper breakdown of what the mismatch actually consists of.

## Break "Assessed" into clickable CoL Match / No CoL Match, drop the warning icon (requested)
The warning icon told you *that* IUCN and CoL disagreed but not *which* species — and investigating it directly (Wild Pig SG: 18 assessed, 17 in CoL) turned up a real modeling bug in `trueAssessed` itself: two assessed species can share one CoL id (a genuine taxonomic lump — *Sus bucculentus*, IUCN Extinct, is CoL-synonymized into *Sus scrofa*, IUCN Least Concern) or one assessed species can carry two CoL links (an `iucn_synonym_covered` bookkeeping alias — a mechanism to stop an already-assessed species resurfacing as a new NE candidate under an alternate CoL name, not a real second species). Naively joining `species_link` conflated both into the aggregate counts. Replaced the flag with a real nested breakdown instead:
```
Suidae (17)
 > Assessed (18)
    CoL Match (17)
    No CoL Match (1)     <- clicking shows Sus bucculentus (EX) directly
 > Not Evaluated (0)
```
- `colBreakdown[].noMatchIds`: the specific `sis_taxon_id`s of assessed species without a clean 1:1 CoL match, computed via a `ROW_NUMBER()` tie-break per `col_id` (preferring an accepted-name link over a synonym-derived one, so *Sus scrofa*'s direct "accepted" link wins and *Sus bucculentus*'s "synonym" link loses) — and deliberately excluding `iucn_synonym_covered` rows from that tie-break, since including them let one assessed id win two different `col_id`s at once and silently undercount the real mismatch (caught by verifying Primate SG's numbers didn't add up either).
- The `bd=` URL param gained an optional `:only:id1,id2` / `:excl:id1,id2` suffix, so clicking "CoL Match" or "No CoL Match" narrows the already-fetched species list to (or away from) that exact id set — no new API endpoint.
- No warning flag needed: the CoL Match/No CoL Match split makes the discrepancy self-evident structurally, and clicking "No CoL Match" takes you straight to the actual species instead of just a number.

## Popover polish (user-reported)
- **Assessed no longer needs a click to expand**: CoL Match/No CoL Match now render immediately whenever there's a mismatch — the chevron/toggle is gone from that row (the top-level name row above it still toggles, since it can be long and most rows have nothing to show).
- **Exclude clause moved below the breakdown list**: a node like Antelope SG (`excludeGenera` of 15 Caprinae/Bison/Wild-Cattle genera) used to show "(excluding Bos, Bubalus, …)" *above* "Family: ▸ Bovidae (217)" with nothing to attach to, once the primary dimension's inline text was hidden in favor of the breakdown list. Now it reads naturally as a qualifier underneath: `Family: ▸ Bovidae (217)` then `(excluding Bos, Bubalus, …)`. Extracted a small `renderFilterSegs()` helper so the segment-rendering logic isn't duplicated between the two possible render positions.
- Investigated a reported case directly rather than assuming it was a bug: *Microtus obscurus* showing as "No CoL Match" under Small Mammal SG is correct — it and *Microtus mystacinus* both link to the same CoL id as *Microtus arvalis* (a CoL synonymy), while IUCN assesses all three as distinct LC species. Same lump pattern as the Wild Pig case, working as designed.

## Explain each "No CoL Match" species inline (requested)
Clicking through to a "No CoL Match" species list to figure out *why* a species didn't match got old fast — added `colBreakdown[].noMatchDetails`: for each mismatched species, a classified reason plus its name, shown directly in the popover:
```
No CoL Match (1)
  Sus bucculentus — Catalogue of Life treats this as the same species as Sus scrofa
```
Reasons: `lumped` (shares a CoL id with another assessed species — shows which one), `not_in_base` (real CoL match, not yet in the curated Base checklist), `missing_from_backbone` (linked CoL id isn't in the current build — e.g. a curated-checklist demotion), `extinct_unconfirmed` (CoL flags it extinct, IUCN hasn't confirmed EX/EW), `no_link`, `classified_elsewhere`. Computed in the same build-time query that already produces `noMatchIds` — one extra diagnostic pass per breakdown name, only when there's something to explain.

Built deliberately modular: `noMatchIds` and the CoL Match/No CoL Match navigation are completely untouched, so this is revertible independently by dropping `noMatchDetails` alone. (Superseded below: the inline capped list this section originally shipped with is now the full paginated panel instead.)

## Paginated species-level panel beside the popup, species open in a new tab (requested)
Clicking a count row used to navigate away from the current page (replacing whatever taxa/filters you had set up) just to see 5-10 species. Now it opens a `SpeciesListPanel` *beside* the popup instead — the popup goes back to counts-only, and clicking any count (Assessed / Not Evaluated / CoL Match / No CoL Match) opens the actual species list in a side panel, paginated 10 at a time, without touching the page underneath:
```
[popup: counts]              [panel: species]
Suidae (17)                  Suidae — No CoL Match
 > Assessed (18)             Open full list ↗
    CoL Match (17)           Sus bucculentus — Catalogue of
    No CoL Match (1)         Life treats this as the same
 > Not Evaluated (0)         species as Sus scrofa
                              ‹ Prev   1-1 of 1   Next ›
```
- Reuses the existing `/api/redlist/species` endpoint (the same broad per-node fetch `RedListView` itself does) — no new API surface. Narrows client-side with the same `speciesMatchesNode`/`matchesBreakdownName` functions `RedListView` uses, then further splits by bucket (CoL Match excludes `noMatchIds`, No CoL Match includes only them). Cached per `(nodeId, assessed|NE)` for the popover's lifetime, so switching between Assessed/CoL Match/No CoL Match — all drawn from the same assessed fetch — never re-fetches.
- Every species name is a real `<a target="_blank">`, not the app's usual pushState navigation — confirmed the *original* tab's URL is completely untouched after opening a species in a new tab, so exploring a discrepancy never costs you your place.
- The "No CoL Match" reason annotations (lumped-with/not-in-base/etc.) moved from the capped 6-item popup list into this panel, uncapped and paginated — verified against Small Mammal SG's Rodentia (86 Not Evaluated species; page 1 showed 1-10, Next correctly advanced to 11-20).
- Added an "Open full list ↗" escape hatch at the top of the panel to `RedListView`'s complete filterable experience (sorting, charts, map, exact filters) — also a new tab — for anyone who wants more than the panel's lighter-weight list.
- Panel positions itself beside the popup (right if there's room, else left, else below) and closes whenever the popup does; clicks inside it don't propagate to the popup's outside-click-to-close listener (it's a portal sibling, not a DOM descendant, so this needed an explicit `onMouseDown` stop rather than relying on the existing ref-containment check).

## Panel polish (user-reported)
- **The "lumped with" species name is now clickable too**: `noMatchDetails`' `lumped` reason gained `detailId` (the winning species' own assessed id — already available from the same `candidate_links`/`winners` CTE that produces its name, just wasn't selected before), so "Sus bucculentus — ... same species as **Sus scrofa**" links to *Sus scrofa*'s own detail row as well, not just the mismatched species.
- **"Open full list ↗" moved into the header row**, before the × close button, instead of sitting on its own line below the title.
- **Fixed a visible gap between the popup and the panel**: position was computed assuming the popup's `max-w-[340px]` as its actual rendered width, but the popup is only ever as wide as its content — a real popup measured ~278px. Now reads the popup's actual `getBoundingClientRect()` instead of assuming the max-width; verified the gap is exactly the intended 8px (`PANEL_GAP`), not the ~70px stray gap a max-width assumption left for a narrower popup.

**Also investigated directly** (dispatched a subagent, not a guess): Antelope SG's 44.2% assessed ratio — the lowest of all 36 groups — turned out to be the exact same lump pattern as Wild Pig SG, just at ~60x the scale. ~90 of its 119 "unassessed" species are concentrated in 10 genera (*Gazella*, *Cephalophus*/*Philantomba* duikers, *Oreotragus* klipspringer, *Tragelaphus* bushbuck, *Damaliscus* topi, *Alcelaphus* hartebeest, *Kobus*, *Madoqua*, *Redunca*) where Catalogue of Life has elevated geographic subspecies of already-assessed IUCN species to full species status. Filter mechanics checked out clean (no excluded-genus leakage, no family contamination) — this is a genuine, well-characterized signal, not a bug. Not yet added: a node-specific tooltip note for `ssc-antelope` explaining this (same mechanism as the existing Bison SG note) — flagged as a good next addition, not done in this pass.

## "Likely split from X" for Not Evaluated species, and fixed the popup running off-screen (requested)
Investigated feasibility first rather than guessing: Catalogue of Life keeps a subspecies-rank synonym record when a subspecies is promoted to full species status (e.g. *Gazella gazella acaciae* survives as a synonym once *Gazella acaciae* becomes its own accepted species) — so a Not Evaluated species whose former name (as a subspecies synonym) matches an IUCN-assessed species' binomial is very likely a recent split off that species. Validated against two independent real cases before building anything: *Gazella acaciae* ← *Gazella gazella* (EN), *Oreotragus aceratos* ← *Oreotragus oreotragus* (LC).

Implemented as `colBreakdown[].splitDetails` — a once-per-build lookup (`SPLIT_CANDIDATES_SQL` precomputes a col_id → likely-former-parent table by joining `backbone.parquet`'s subspecies/infraspecific synonyms to `assessed.parquet` by binomial, so per-name breakdown queries don't re-scan the ~3.8M-row backbone each time), keyed by `col_id` since Not Evaluated species have no `sis_taxon_id`. Rendered inline in the species panel's Not Evaluated rows as "likely split from **X**" (X clickable), same additive/droppable pattern as the existing "lumped with" reason. Deliberately worded as a heuristic ("likely"), not fact — it only catches the "was a named subspecies, got promoted" pattern (the common case in recent Bovidae splits) and won't catch a split into a segregate with no prior CoL subspecies record.

**Also fixed**: the "# Described Species" popup (and its species-list panel) could extend below the bottom of the viewport when opened near the bottom of the screen, making the lower part of the popup completely inaccessible — it only ever clamped horizontal position, and its `max-h-[70vh]` was a fixed height regardless of how little room was actually left below the trigger button. Both `computePopoverPos` and `computePanelPos` now derive max-height from actual space around the trigger, and flip to open upward when there isn't enough room below.

## Popup regressions fixed (user-reported)
Two real bugs surfaced immediately after the changes above shipped:
- **Scrolling inside the popup/panel closed the whole thing.** The outside-close listener ran on `scroll`'s capture phase with no containment check at all (unlike the `mousedown` listener, which already exempted clicks inside the popover) — so scrolling the popup's or panel's own `overflow-y-auto` content (which fires a `scroll` event captured by the same document-level listener, since capture-phase listeners on an ancestor still fire for a target's own non-bubbling scroll event) closed everything before you could read past the fold. This was the real cause behind the previous "popup runs off-screen" report too — the max-height clamp made scrolling *inside* the popup necessary far more often, which is what exposed this. Fixed by exempting scroll (and mousedown) events whose target is inside the popover, the panel, or the trigger button — a real page/table scroll still closes it as before.
- **Popup shouldn't flip upward, ever.** Simplified `computePopoverPos`/`computePanelPos` to always open downward/beside the trigger and never reposition to fit — they only clamp max-height to whatever space is actually available, leaving the (now-fixed) internal scroll to handle anything that doesn't fit. Position relative to the row that triggered it is always predictable now.

**Also found and fixed a real gap in the "split from" heuristic** (asked directly: why isn't *Giraffa tippelskirchi* flagged as split from *Giraffa camelopardalis*?). Checked the backbone chain directly: CoL doesn't always point the old subspecies synonym straight at the promoted species — when the promoted species also got its own newly-created nominate subspecies (an "autonym", e.g. *Giraffa tippelskirchi tippelskirchi*), the old synonym (*Giraffa camelopardalis tippelskirchi*) points at that nominate **subspecies**, not at the species itself, so the original single-hop lookup missed it entirely. Now resolves one extra hop whenever the synonym's immediate parent is itself an accepted subspecies/infraspecific rank. Fixes all three *Giraffa* species (*giraffa*, *reticulata*, *tippelskirchi*) to correctly show "split from *Giraffa camelopardalis*" — verified in the rebuilt data and in the browser.

## Stop closing on scroll entirely (requested)
The scroll-close listener (kept, just fixed to exempt the popover's own content, in the section above) turned out to be the wrong behavior altogether: the request was to let the popup/panel stay open while scrolling the *rest* of the page, closing only on an actual outside click. Removed the scroll listener entirely — `position: fixed` already keeps both correctly anchored to the viewport regardless of what scrolls underneath, so there was never a real staleness problem to guard against. Outside-click and Escape still close them as before.

## "missing_from_backbone" was actively wrong, not just confusing (user-reported)
Asked directly: what does "its Catalogue of Life match isn't in the current backbone" mean for a case like *Arctocephalus townsendi*, and is it correct? Investigated with real `duckdb` queries rather than just rewording the label. *Arctocephalus townsendi* (Guadalupe fur seal, IUCN LC) is currently filed by Catalogue of Life as *Arctocephalus philippii townsendi* — a **subspecies** of the assessed *Arctocephalus philippii* (Juan Fernández fur seal, LC), not its own accepted species. That part is a genuine, correctly-detected taxonomic disagreement between CoL and IUCN — the same shape as the existing "lumped" reason, just via subspecies-demotion instead of a synonym tie-break.

But the label itself turned out to be wrong, not just unclear to a non-technical reader: checked **every** case hitting `missing_from_backbone` across the whole dataset (1,198 of them) — **zero** are actually missing from `backbone.parquet`. Every one resolves to a real backbone row that's either a subspecies/variety/genus-rank record (91%, CoL just doesn't recognize it as its own species) or a species-rank record that's "provisionally accepted" rather than "accepted" (9%, it exists, just isn't in CoL's curated checklist yet). Split into two accurate reasons:
- **`infraspecific`** (the majority case, like *Arctocephalus townsendi*): "Catalogue of Life doesn't recognize this as a distinct species — it's currently classified as part of **Arctocephalus philippii**", the parent linked via a new `col_to_assessed` precomputed lookup whenever it's itself IUCN-assessed — same `detail`/`detailId` pattern the "lumped" reason already uses.
- **`provisional`**: "matched to a Catalogue of Life name that's only provisionally accepted, not yet fully accepted" — distinct, less alarming wording for the smaller, genuinely different case.

`missing_from_backbone` is kept only for the genuine dangling-reference case (structurally possible, just never actually observed) and as a fallback when `backbone.parquet` isn't available at build time.

## Recent-description note for unexplained NE species, and fixing another misleading label (requested)
Spotted directly by the user: *Manis mysteria* and *Manis indoburmanica* (both Pangolin SG) had no explanation for Not Evaluated status, but are both very recently described. Confirmed against real data (`described_year` in `species/`): *Manis mysteria* = 2023, *Manis indoburmanica* = 2025. That field is already fetched client-side for every NE row (`RedListSpecies.described_year`), so this needed no build-script or data-pipeline change — a pure frontend addition. NE rows with no split-from match now show `(described in <yr>)` when described within the last 10 years (mirrors the existing "# Outdated (>10 yrs old)" threshold elsewhere in the dashboard) — a genuine, likely explanation (assessment backlog), deliberately not shown for older description years since those don't reliably explain anything.

Also asked directly: is "CoL Match"/"No CoL Match" still accurate given the infraspecific/provisional split above? Checked: of everything currently bucketed under "No CoL Match," only ~3% (the `no_link` reason) genuinely has zero Catalogue of Life record — the other ~97% (lumped, infraspecific, provisional, not_in_base, extinct_unconfirmed, classified_elsewhere) all have a real CoL record, just not one that counts as a *clean, single, species-level* match. Renamed to **"Clean CoL Match"** / **"No Clean CoL Match"** and updated the tooltip to say so explicitly, rather than implying CoL has no knowledge of ~97% of these species when it usually does.

## Panel upgrade: 1:1 CoL Match rename, category/date, search, sort (requested)
Renamed the positive side of the match split from "Clean CoL Match" to **"1:1 CoL Match"** — more concrete about what's actually guaranteed (a single, unambiguous correspondence), while keeping "No Clean CoL Match" as-is on the negative side (considered "No CoL Species Match" too, but rejected it: it reads as "not a species in CoL," which is only true for the small `no_link` minority — most of the negative bucket *does* have a real CoL species record, just not one that counts as a clean match).

Then, three panel additions, all pure frontend since the underlying data (`category`, `assessment_date`, `common_name`) is already fetched per species row — no build/data change:
- Each row now shows its RL category (color-coded, same palette as the rest of the dashboard) and assessment year.
- A search box, matching scientific *or* common name (verified: searching "rat" correctly surfaces Rodentia's several "chinchilla rat" species by common name, not just literal scientific-name substring hits).
- A Name/Date sort toggle. For the NE bucket, "Date" sorts by description year instead (NE rows have no assessment date) — pairs directly with the "(described in \<yr\>)" note added earlier, so the most-recently-described (most-likely-just-not-caught-up-yet) species surface first.

Along the way, consolidated the panel's pagination/search/sort reset into one effect keyed on the full view identity (`nodeId` + `bucket` + `name` + `rank`) instead of the pagination-only reset that was keyed on just `[nodeId, isNe]` — the old keying meant switching between two non-NE buckets (e.g. 1:1 Match ↔ No Clean Match, both `isNe = false`) never fired the reset, a latent stale-pagination bug now also fixed for search/sort.

## Scroll-tracking popup, spinner, more label/sort polish (requested)
- **Popup and panel now track the row while the page scrolls**, instead of staying frozen at their original screen position (the prior "don't close on scroll" fix stopped them from disappearing, but left them drifting away from the row they were opened from as you scrolled past it). A capture-phase scroll listener recomputes position on scroll, rAF-throttled, and explicitly ignores scroll events whose target is inside the popup/panel's own content (so scrolling a long species list doesn't drag the whole popup along with it). Verified with real bounding-rect measurements: the offset between the trigger icon and the popup stayed exactly constant (14px) across a 300px scroll, and the panel similarly stayed pinned to the popup across a 200px scroll.
- Removed the "sort by description year" option for the NE bucket — only Name sort is offered there now (the per-species recency note already covers that signal; a redundant sort control wasn't wanted).
- **"No Clean CoL Match" → "No 1:1 CoL Match"**, for full symmetry with "1:1 CoL Match".
- Sort button label **"Date" → "Assess. Yr"**.
- **Loading spinner** instead of a plain "Loading…" text label, matching the SVG spinner already used elsewhere in this file.

**Also investigated directly** (not a code change — a genuine heuristic limitation, not a bug): why don't *Alcelaphus lichtensteinii* and *Alcelaphus caama* get a "likely split from" annotation, when their Bovidae relatives (*A. cokii*, *A. lelwel*, *A. major*, *A. swaynei*, *A. tora*) do? Checked the backbone directly: those five all have a retained trinomial synonym record (e.g. "Alcelaphus buselaphus cokii") pointing straight at their promoted species — exactly the pattern the heuristic looks for. *lichtensteinii* and *caama* have **zero** subspecies-rank synonyms in Catalogue of Life's data at all; their only synonyms are unrelated species-rank basionyms (*Antilope caama*, *Antilope lichtensteinii*, *Sigmoceros lichtensteinii*) — CoL's own checklist simply never modeled them as former subspecies of *Alcelaphus buselaphus*, unlike their five siblings. This is precisely the documented caveat ("won't catch a split into a segregate with no prior CoL subspecies record") — correct, expected behavior, not fixable without a broader (and riskier, false-positive-prone) heuristic.

## Test plan
- [x] `npx tsc --noEmit` — clean
- [x] `npm run lint` — clean
- [x] `npm test` — 489/489 passing, including tests for the genus/species filter logic, the trickier SSC boundary splits, the URL round-trip regression, and the remainder row's non-overlap with all 35 named groups
- [x] Visual browser check (Playwright + Node 22 LTS — Node 26's zip extraction hangs, unrelated to this repo): landing page → SSC groups mode → full table with real numbers, sorted by assessed descending → drill into a group and into the remainder row → correctly-filtered species lists with working charts/map → click Mammals breadcrumb, returns to SSC groups mode
- [x] Mobile (390px) + desktop (1400px) visual check: toggles reachable and wrap sensibly on mobile; desktop table has no horizontal overflow
- [x] Verified against a full production build (`npm run build && npm start`), not just `next dev`
- [x] Verified the CoL-default switch directly against the backbone with `duckdb` CLI queries (not just app-level testing) for every claimed fix — Bison, the 7 domestic exclusions, the Hippo gap
- [x] Verified the described-species popover across cases: a plain CoL-sourced SSC group (Cat SG), a group with a CoL override note (Bison SG), a now-visible CoL-undercount case (Hippo SG, showing 400% assessed with an explanatory note), and an IUCN-cited row (Mammals, Bats) — each shows the correct explanation/links; confirmed `build-taxa-summary.ts`'s refactor to import the shared override config produces byte-identical data output
- [x] Verified the popover survives mouse movement onto it (the bug this replaced) and that its links resolve to the correct CoL taxon/search pages (spot-checked several URLs against the source name)
- [x] Verified `build-col-taxon-ids.ts`'s resolution rate (167/177) and spot-checked several of the 10 unresolved names directly against `backbone.parquet` to confirm they're genuine CoL classification differences, not bugs
- [x] Verified per-name breakdown counts sum to `colDescribed` exactly (`duckdb` CLI, cross-checked against Pinniped SG and Small Carnivore SG) and confirmed in the browser (Pinniped SG: "Otariidae (14), Phocidae (18), Odobenidae (1)")
- [ ] **Not verified**: whether the minimal CoL search URL (dropped `facet=…` params) actually returns the same result set as the full copy-pasted one — blocked by a 403 when automating a real browser against catalogueoflife.org; needs a manual click-through
- [x] Verified the CoL refresh directly against ChecklistBank's live API (`api.checklistbank.org` — no browser needed, no bot-blocking): confirmed `fetch-coldp.ts` now resolves the current XR dataset (`315557`/`COL26.6`) instead of the stale hardcoded one, confirmed Rodentia/Suidae counts now match the live site exactly, confirmed `Choeropsis liberiensis`'s accepted/synonym status flip between releases
- [x] Verified the extinct/EX-EW universe rule with real queries: the 222-species corroborated set (spot-checked several — Great Auk, Thylacine, Mauritian giant tortoises), the mismatch cases it correctly excludes (`Columba arquatrix` and others linked to non-EX assessments), and Hippo SG's exact 4/4 in both the summary table and the live "Unassessed" species-list toggle (0 species, confirming no stray never-assessed fossil leaks through)
- [x] Verified the client-side/server-side matching fix: Cat SG's count and its displayed "Unassessed" species list now both agree on 6 species (no `Felis catus`); confirmed via `species_link` that `Leopardus munoai`'s absence from the list is a real assessment gap, not a matching bug
- [x] Verified the Cetacean SG fix in the browser: popover breakdown now shows `Hyperoodontidae (24)` and `Neobalaenidae (1)` as working CoL links (`Ziphiidae (0)` correctly renders as plain text, unresolved), summing to the full 96; confirmed the tooltip link's href no longer includes `provisionally accepted`
- [x] Verified the Assessed/Unassessed toggle fix in the browser: drilled into Mammals, toggled to Unassessed — URL and table stayed scoped to Mammals (534 species) instead of resetting to the "All Species" landing page; toggled back, taxa selection still intact
- [x] Verified the SSC-subgroup-scope fix in the browser: drilled into Cetacean SG (`?taxa=ssc-cetacean`), toggled to Unassessed — URL stayed `?view=new-assessments&taxa=ssc-cetacean` instead of falling back to `taxa=mammals`
- [x] Verified the "Click this row" text is gone from the popover (button count 0) and the popover's "Source" line reads "Catalogue of Life (COL26.6 XR)" linking to `https://doi.org/10.48580/dgy8b` — confirmed for both the per-row popover and the column-header info icon
- [x] Re-ran `build-col-taxon-ids.ts` + `build-taxa-summary.ts` after the Cetacean fix, `npm run diff-data-vs-r2` confirmed only `node-children-summaries.json` changed (156 bytes), and published via `npm run upload-data-to-r2`
- [x] Verified the expandable breakdown in the browser: Small Mammal SG's popover shows `Order:` as a list; expanding Rodentia shows `Assessed (2302)` / `Not Evaluated (86)`, summing to `2388`; clicking "Not Evaluated" navigates to `?view=new-assessments&taxa=ssc-small-mammal` and shows all 160 NE species (Rodentia + Eulipotyphla + Scandentia combined, as documented — not narrowed to Rodentia alone)
- [x] Re-ran `build-taxa-summary.ts` for the new per-name `neCount`, `npm run diff-data-vs-r2` confirmed only `node-children-summaries.json` changed (+5,138 bytes, all `neCount` additions), and published via `npm run upload-data-to-r2`
- [x] Verified the name-narrowing in the browser end to end: clicking Rodentia's "Not Evaluated" in Small Mammal SG's popover shows exactly 86 species with a "Rodentia ×" chip; clearing the chip restores all 160; loading a URL with a `bd=` for one group then clicking a *different* SSC group's row in-app (Cat SG) correctly drops it and shows Cat SG's normal 39 species, not an empty/filtered list
- [x] Caught and fixed the setSelectedTaxa/view-mode-switch race described above by instrumenting `syncUrl` and `navigateToNodeSpeciesList` with temporary logging, tracing the exact sequence of URL writes after a popover click, then removed the logging once confirmed fixed
- [x] Verified single-name breakdown + mismatch flag directly against the rebuilt data (`node-children-summaries.json`): Primate SG → `{count:527, neCount:6, trueAssessed:527}`; Wild Pig SG → `{count:17, neCount:0, trueAssessed:18}`; confirmed in the browser — Primate SG's popover now shows an expandable `Primates (527) ⚠`, Wild Pig SG's shows `Suidae (17) ⚠` expanding to `Assessed (17) (IUCN: 18)` / `Not Evaluated (0)`
- [x] Re-ran `build-taxa-summary.ts` (242 per-name breakdown queries now, up from 202 — every node with a resolvable primary dimension, not just multi-name ones), `npm run diff-data-vs-r2` confirmed only `node-children-summaries.json` changed (+12,379 bytes), and published via `npm run upload-data-to-r2`
- [x] Verified the CoL Match/No CoL Match split directly against the rebuilt data: Wild Pig SG's `suidae` entry → `{trueAssessed:18, noMatchIds:[21178]}` (21178 = *Sus bucculentus*, confirmed via `assessed.parquet` and cross-checked against `species_link.parquet` that it and *Sus scrofa* (id 41775) share `col_id` 53HGR, one via `match_method='accepted'` and one via `'synonym'`)
- [x] Verified in the browser end to end: Wild Pig SG's popover → `Suidae (17)` → `Assessed (18)` → `CoL Match (17)` / `No CoL Match (1)`; clicked "No CoL Match" and confirmed it navigates straight to a 1-species list showing exactly *Sus bucculentus* (EX), with a "Suidae — No CoL Match ×" chip
- [x] Caught and fixed a real bug in `trueAssessed`/`noMatchIds` before shipping: an initial version double-counted assessed species with an `iucn_synonym_covered` link (letting one id "win" two `col_id`s), which a build-time sanity check (`trueAssessed - noMatchIds.length` should equal the CoL-side linked count) flagged for ~50 nodes including Primate SG; root-caused and fixed by excluding covered-link rows from the tie-break, re-verified Wild Pig SG and Primate SG by hand afterward
- [x] Re-ran `build-taxa-summary.ts` after the fix, `npm run diff-data-vs-r2` confirmed only `node-children-summaries.json` changed (+32,781 bytes), and published via `npm run upload-data-to-r2`
- [x] Verified the popover polish in the browser: Wild Pig SG's "Assessed (18)" now shows "CoL Match (17)" / "No CoL Match (1)" without any click; Antelope SG's popover reads "Family: Bovidae (217)" then "(excluding Bos, Bubalus, …)" below it, not above
- [x] Verified the *Microtus obscurus* report directly against `species_link.parquet`/`assessed.parquet`: confirmed it's a genuine CoL synonymy (shares `col_id` 7VV7P with *Microtus mystacinus* and *Microtus arvalis*), not a matching bug — no code change needed for that part
- [x] Verified `noMatchDetails` directly against the rebuilt data: Wild Pig SG → `{id:21178, name:"Sus bucculentus", reason:"lumped", detail:"Sus scrofa"}`; Antelope SG → `{id:22052, name:"Tragelaphus angasii", reason:"not_in_base"}` (matches the earlier manual audit exactly); Primate SG's 9 mismatches span `lumped`/`not_in_base`/`missing_from_backbone`, all consistent with prior findings
- [x] Verified in the browser: Wild Pig SG's popover shows "Sus bucculentus — Catalogue of Life treats this as the same species as Sus scrofa" inline; clicking the species name navigates to `?taxa=ssc-wild-pig&species=21178` and lands directly on its expanded detail row in the full 18-species list
- [x] Checked the largest `noMatchDetails` case (squamates/Squamata: 152) to confirm the 6-item cap + "+N more" fallback behaves sensibly rather than dumping the whole list
- [x] Dispatched an independent subagent to verify Antelope SG's 44.2% figure with real `duckdb` queries rather than assuming it was a bug — see the section above
- [x] Re-ran `build-taxa-summary.ts` (15.3s, 242 breakdown queries, one extra diagnostic pass per name with a mismatch), `npm run diff-data-vs-r2` confirmed only `node-children-summaries.json` changed (+191,756 bytes, expected — the largest mismatched groups, e.g. Rodentia's ~140 entries, now carry full name+reason detail), and published via `npm run upload-data-to-r2`
- [x] Verified the species-level panel in the browser end to end: opened it for Wild Pig SG's "No CoL Match" — popup and panel both stayed open, page URL unchanged (confirmed via automated URL check, not just visual); clicked the species link inside the panel — a genuine new browser tab opened at `?taxa=ssc-wild-pig&species=21178`, landed directly on that species' expanded row, and the original tab's URL was still unchanged afterward
- [x] Verified pagination on a large bucket (Small Mammal SG's Rodentia, Not Evaluated, 86 species): panel showed "1-10 of 86"; clicking Next correctly advanced to a different set of 10 ("11-20 of 86") — no code change needed, this is purely a no-new-data-required UI feature (no R2 publish for this commit)
- [x] Verified the panel polish in the browser: measured actual bounding boxes for Wild Pig SG's popup (278px rendered width) and panel — gap is exactly 8px; confirmed `Sus scrofa` renders as a link with `href="/?taxa=ssc-wild-pig&species=41775"`; confirmed "Open full list ↗" appears in the header row alongside the × button
- [x] Re-ran `build-taxa-summary.ts` for the new `detailId` field, `npm run diff-data-vs-r2` confirmed only `node-children-summaries.json` changed (+13,362 bytes), and published via `npm run upload-data-to-r2`
- [x] Validated the "split from" heuristic against real data *before* implementing: `duckdb` CLI queries confirmed both *Gazella acaciae*→*Gazella gazella* and *Oreotragus aceratos*→*Oreotragus oreotragus* independently, then re-confirmed both survived in the rebuilt data after implementation (`splitDetails` entries: `{colId:"VBLG2", parentId:8989, parentName:"Gazella gazella", parentCategory:"EN"}`, `{colId:"RBJ76", parentId:15485, parentName:"Oreotragus oreotragus", parentCategory:"LC"}`)
- [x] Re-ran `build-taxa-summary.ts` (17.2s, no meaningful slowdown from the new once-per-build precompute), `npm run diff-data-vs-r2` confirmed only `node-children-summaries.json` changed (+157,374 bytes)
- [x] Verified in the browser: Antelope SG's Bovidae → Not Evaluated panel shows "Oreotragus transvaalensis — likely split from Oreotragus oreotragus", "Tragelaphus phalerates — likely split from Tragelaphus scriptus", and several more, each parent name a working link
- [x] Verified the popup-clamp fix with real bounding-rect measurements, not just visual inspection: opened the popup with its trigger button `scrollIntoView`'d to the very bottom edge of a 700px-tall viewport — popup flipped to open upward and its full rect stayed within `[0, 700]`, where the old fixed-`top`/`max-h-[70vh]` logic would have placed its bottom well past 1000px
- [x] Published to R2 (sync `2026-07-10T10-57-44Z`) after explicit confirmation, same pattern as every prior data change in this PR
- [x] Verified the scroll-close fix in the browser with a real mouse-wheel scroll over the popup/panel content (Playwright `mouse.wheel`), not just a click: panel stayed open and its content was still visible afterward
- [x] Verified the popup never flips upward: repeated the same worst-case check as before (trigger `scrollIntoView`'d to the bottom edge of a 700px viewport) — popup now opens *below* the button every time (`popup.top > icon.top`), instead of flipping above it
- [x] Verified the Giraffa two-hop fix directly against the backbone before implementing (`duckdb` CLI: confirmed *Giraffa camelopardalis tippelskirchi*'s synonym record points at the new nominate subspecies *Giraffa tippelskirchi tippelskirchi*, not at the species), then re-verified after implementing — `splitDetails` for all three *Giraffa* species now correctly resolve to `Giraffa camelopardalis` (id 9194, VU); confirmed in the browser (Giraffe and Okapi SG → Giraffidae → Not Evaluated panel)
- [x] Re-ran `build-taxa-summary.ts` (17.6s), `npm run diff-data-vs-r2` confirmed only `node-children-summaries.json` changed (+7,882 bytes), and published via `npm run upload-data-to-r2` (sync `2026-07-10T11-12-44Z`)
- [x] Verified the no-scroll-close change in the browser (DOM-node check, not a text-substring check — an earlier pass in this same verification gave a false negative because "Catalogue of Life" text also appears elsewhere on the page, e.g. the source link, independent of whether the popup itself is open): popup stayed open through a window-level mouse-wheel scroll and a table-body horizontal scroll, then closed correctly on an outside click. No data change, no R2 publish needed for this commit.
- [x] Verified the infraspecific/provisional split directly against real data before implementing: `duckdb` confirmed all 1,198 `missing_from_backbone` cases resolve to real backbone rows (0 genuinely dangling), broken down 91% infraspecific-rank / 9% provisionally-accepted; re-verified after implementing — Pinniped SG's *Arctocephalus townsendi* → `{reason:"infraspecific", detail:"Arctocephalus philippii", detailId:2059}`; confirmed in the browser (clickable parent link, correct wording)
- [x] Re-ran `build-taxa-summary.ts` (31.6s, up from 17.6s — the extra backbone joins roughly double per-breakdown query cost, still acceptable for a build step), `npm run diff-data-vs-r2` confirmed only `node-children-summaries.json` changed (+6,983 bytes), and published via `npm run upload-data-to-r2` (sync `2026-07-10T11-36-04Z`)
- [x] Verified the described-year note in the browser: Pangolin SG's Pholidota → Not Evaluated panel shows "Manis mysteria (described in 2023)" and "Manis indoburmanica (described in 2025)"
- [x] Verified the Clean CoL Match rename in the browser: Wild Pig SG's Suidae breakdown now reads "Clean CoL Match (17)" / "No Clean CoL Match (1)", tooltip updated. No data change, no R2 publish needed for this commit (both changes read from data already fetched client-side).
- [x] Verified the panel upgrade in the browser end to end (Small Mammal SG's Rodentia, 2,244-species "1:1 CoL Match" bucket): category badges (DD/LC/CR/VU/EN, color-coded) and assessment years render per row; searching "rat" correctly narrows to 492 species matching by common name (chinchilla rats etc.), not just scientific-name substrings; clicking "Date" sorts newest-assessed-first (2025 entries surfaced immediately), clicking again reverses to oldest-first; label now reads "1:1 CoL Match" throughout (header, button, panel title). No data change, no R2 publish needed for this commit.
- [x] Verified scroll-tracking with real bounding-rect measurements (not just visual inspection): opened Antelope SG's popup, measured the icon-to-popup vertical offset (14px), scrolled the page 300px, re-measured — offset unchanged (still 14px), popup moved by ~300px along with the row; repeated for the species panel relative to the popup across a 200px scroll — same result
- [x] Verified in the browser: Wild Pig SG's Suidae now reads "1:1 CoL Match (17)" / "No 1:1 CoL Match (1)"; Pangolin SG's Not Evaluated panel (throttled network, to actually observe the loading state) shows the spinner instead of "Loading…" text, then correctly hides the "Assess. Yr" sort button (NE bucket) while keeping "Name" and the search box
- [x] Investigated the Alcelaphus split-from gap directly against the backbone (`duckdb`): confirmed *A. cokii*/*lelwel*/*major*/*swaynei*/*tora* all retain a trinomial synonym pointing at their promoted species, while *A. lichtensteinii*/*caama* have zero subspecies-rank synonyms recorded at all — a genuine data gap in Catalogue of Life's own checklist, correctly and expectedly not caught by the heuristic. No code change from this investigation.

## Next steps
- [ ] Manually spot-check the minimal CoL search URL (e.g. click through on Pinniped SG's Otariidae link) — confirm it shows the same result set as the full facet-laden URL, since this couldn't be automated-tested (see Test plan).
- [ ] Re-verify the Vercel preview deployment now that the sync pointer has moved again — confirm real data shows there before merging.
- [ ] Merge once satisfied.
- [ ] Optional/future: if this pilot lands well, extend SSC groups mode to non-mammal taxa (birds, reptiles, etc.) — similar-sized research + build effort per taxon group.


shaneweisz:
Found and fixed the deployment failure: it wasn't a code bug. \`npm run build\`'s \`prebuild\` step (\`fetch-data-from-r2\`) unconditionally re-downloads every file from the R2 sync pinned in \`app/latest-sync.txt\`, which was overwriting the locally-regenerated \`data/node-children-summaries.json\` (the one with the new \`ssc-groups\` entry) with the stale pre-PR copy — so the SSC groups endpoint came back empty on the deployed build.

Published a new R2 sync (\`2026-07-07T13-37-48Z\`, uploaded 123 files) with the regenerated data and bumped \`app/latest-sync.txt\` — the new build should now show real per-group numbers.

shaneweisz:
Root-caused the Playwright hang too, for the record: it wasn't the sandbox network (proved fine with plain \`curl\`/\`fetch\` — same 162MB file in 5s) — Playwright's in-process zip extraction was deadlocking specifically on Node v26.4.0 (very new, likely untested against Playwright yet). Installed Node 22 LTS side-by-side (keg-only via brew, doesn't touch the default \`node\`) and extraction worked cleanly.

That let me actually drive the feature in a real headless browser end-to-end:
- Landing page → "SSC groups mode" button renders correctly next to "Table 1a mode"
- Full 35-group table renders with real numbers, subtotal/total rows
- Clicking "Bat Specialist Group" drills into a correctly-filtered species list (1,336 species, all genuinely bats) with working charts/map/threats, URL becomes `?taxa=ssc-bat` as designed

Caught and fixed one real bug this way: the breadcrumb above the species table showed bogus "All Species" / "SSC Specialist Groups" rows before "Bat Specialist Group", because the ancestor-walk was following the real tree parent (`ssc-groups` → `all`) instead of stopping where SSC groups mode navigates as if nested (`mammals`). Fixed in the latest commit — breadcrumb is now clean (`All Species → Mammals → Bat Specialist Group`).

Also confirmed cosmetically: a few groups' `estimatedDescribed` are too conservative vs. real assessed counts (Hippo shows 200%, a couple others >100%) — bars clamp fine, but those specific estimates are worth a follow-up spot-check.

All checks still green after the fix: tsc clean, lint clean, 487/487 tests passing.

shaneweisz:
Added: each SSC group row now has a small ⓘ next to its name — hover shows "View official IUCN page" and links to that specific group's own iucn.org page (the same pages used to source its species filter in the first place).

Verified visually (Node 22 + Playwright): hovered Bat Specialist Group, confirmed the tooltip and that the link resolves to \`iucn.org/our-union/commissions/group/iucn-ssc-bat-specialist-group\`. tsc/lint/tests all still clean. No data republish needed this time — the link is looked up client-side from the tree config, not baked into the precomputed summary JSON.

shaneweisz:
Resolved a merge conflict with \`main\` — an unrelated automated data refresh (PR #358) landed while this was open, bumping the pinned sync to \`2026-07-07T12-22-14Z\` with fresher species data but no \`ssc-groups\` computation.

Didn't trust git's line-level auto-merge of the two precomputed JSON files (real risk of silently blending two inconsistent data snapshots) — instead re-fetched the fresher sync, re-ran \`scripts/build-taxa-summary.ts\` on top of it, and republished as a new sync (\`2026-07-07T14-35-31Z\`). Confirmed via \`diff-data-vs-r2\` the only delta from the fresher base is the \`ssc-groups\` addition — everything else matches \`main\` byte-for-byte. tsc/lint/487 tests all still green after the merge.

shaneweisz:
Two small polish items:
- Simplified the usage hint text: "Click taxa rows to filter, use charts and search to explore species. Cmd/Ctrl+click to multiselect." → "Click to filter, Cmd/Ctrl+click to multi-select."
- The SSC group name column's width cap wasn't fully sticking — \`position: sticky\` + \`table-layout: auto\` let the column render wider than its \`max-width\` in practice (~232px against a 220px cap), enough to occasionally crowd the "# Outdated (>10 yrs old)" header. Added \`overflow-hidden\` alongside \`max-width\` (previously only on row cells, not the header) so the constraint actually holds, and tightened the breakpoints to 120/170/220px (from 160/240/360px). Verified via computed-style + screenshot: no horizontal overflow, full "# Outdated (>10 yrs old)" header visible with ~190px of margin to spare at 1400px. More names now truncate at typical widths (hover tooltip still shows the full name).

tsc/lint/487 tests all still green.

PR #368, opened 2026-07-16
Fix country map name mismatches hiding data for ~40 territories

shaneweisz:
## Summary

Follow-up to Thom's feedback about Somaliland/North Macedonia/Kosovo showing "No data" on the country map. Investigated against the live TopoJSON (`world-atlas@2/countries-50m.json`) and the real species data — it's a real bug, bigger than the 3 territories Thom noticed.

**Root cause:** `WorldMap.tsx`'s TopoJSON-name → alpha-2 lookup table was hand-built and doesn't exactly match the strings this map file actually uses for its shapes (e.g. it has `"North Macedonia"` as a key, but the shape is literally named `"Macedonia"`). When a shape's name isn't found in the table, it silently renders as "No data" regardless of whether the Red List has species there. Diffing the table against all 241 TopoJSON shape names found **~40 territories** affected this way.

**Fixed:**
- North Macedonia — hid **1,649** assessed species (the exact bug Thom flagged). Added a display override so tooltips/chips still read "North Macedonia" even though the lookup key now matches the shape's legacy TopoJSON string.
- 7 more entries existed under a full name that didn't match the shape's abbreviated string (Cabo Verde, São Tomé and Príncipe, St. Kitts and Nevis, St. Vincent and the Grenadines, Vatican City, Marshall Islands, Antigua and Barbuda) — renamed the lookup keys, kept the nice display names.
- ~30 territories had no lookup entry at all (Hong Kong, Macao, US/British Virgin Islands, Guam, American Samoa, Cayman Islands, Bermuda, Guernsey, Jersey, Isle of Man, and more) — added them.

**Somaliland and N. Cyprus — mapped to their parent country, not left blank:**
Both are drawn as their own shape in the TopoJSON, but researched IUCN's own Red List country-code standard (redlist.org/resources/country-codes) rather than guessing: it's ISO 3166-1 + UN country names, with no distinct entry for Somaliland or Northern Cyprus — both fold into Somalia/Cyprus, the same way IUCN's own species assessments do (e.g. the Gerenuk assessment's formal country field lists "Somalia" even though its range-description text separately mentions "Somaliland"). IUCN does create bespoke entries for other disputed territories when it chooses to (Palestine, Western Sahara, a generic "Disputed Territory" entry for the Paracels/Spratlys) — the absence for these two looks deliberate, not an oversight. So both shapes now resolve to their parent's alpha-2 code (with explicit display-name overrides so filter chips/tooltips still read "Somalia"/"Cyprus" either way), rather than leaving a "no data" gap that reads as a bug.

**Kosovo — not a bug, a genuine data gap:**
Checked directly against the raw data: the lookup was already correct (`"Kosovo": "XK"`), and there are genuinely **zero** assessed species tagged to Kosovo (there are 2,647 in the *Unassessed* dataset). This one isn't a dashboard bug — it's a real Red List data-granularity gap, exactly as Thom suspected.

## Test plan

- [x] Typecheck passes (`npx tsc --noEmit`), lint clean
- [x] Programmatically diffed the fixed lookup table against all 241 TopoJSON shape names — confirmed 0 unexpected gaps remain (only 3 genuinely uninhabited/no-ISO-code territories left unmapped: Ashmore and Cartier Is., Indian Ocean Ter., Siachen Glacier)
- [x] Queried the real species data (`assessed.parquet`/`unassessed.parquet` via duckdb) to confirm real, non-trivial species counts exist for each previously-hidden code (e.g. MK: 1,649, HK: 1,921, CV: 1,281, KN: 1,725, VC: 1,928, and more)
- [x] Verified in-browser (Playwright): searching for each fixed territory now successfully selects it (confirms the shape's own name now resolves through the same lookup) and shows the correct display name as a filter chip — checked North Macedonia, Cape Verde, Hong Kong, Vatican City, Marshall Islands, Saint Kitts and Nevis, Kosovo
- [x] Verified Somaliland and N. Cyprus now select/color together with Somalia/Cyprus respectively, with filter chips correctly reading "Somalia"/"Cyprus" (not the breakaway region's name) regardless of which shape was clicked
- [x] Kosovo behaves as expected: selectable (mapping was always correct), but genuinely zero assessed-species data
- [x] Confirmed `lib/countries.ts` (a separate, unrelated free-text country-name resolver used for MCP/agent dashboard links) doesn't share this bug and wasn't touched

Built and tested in an isolated git worktree off `main` since other work was in progress concurrently.

### DIFF
--- a/app/src/components/WorldMap.tsx
+++ b/app/src/components/WorldMap.tsx
@@ -68,11 +68,7 @@ const NAME_TO_ALPHA2: Record<string, string> = {
   "Marshall Is.": "MH",
   // Additional territories present as their own shape in the 50m TopoJSON that
   // had no entry at all before (always rendered as "No data" regardless of the
-  // underlying Red List data). Deliberately NOT included: Somaliland and
-  // N. Cyprus, which the TopoJSON also draws separately, but the Red List data
-  // has no distinct code for either (both fold into Somalia/Cyprus) — coloring
-  // them would misattribute data to a disputed territory, so they're left as
-  // "No data" pending a check with the Red List Unit.
+  // underlying Red List data).
   "American Samoa": "AS", "Anguilla": "AI", "Aruba": "AW", "Bermuda": "BM",
   "Br. Indian Ocean Ter.": "IO", "British Virgin Is.": "VG", "Cayman Is.": "KY",
   "Cook Is.": "CK", "Curaçao": "CW", "Faeroe Is.": "FO", "Fr. Polynesia": "PF",
@@ -83,6 +79,15 @@ const NAME_TO_ALPHA2: Record<string, string> = {
   "St-Barthélemy": "BL", "St-Martin": "MF", "St. Pierre and Miquelon": "PM",
   "Turks and Caicos Is.": "TC", "U.S. Virgin Is.": "VI", "Wallis and Futuna Is.": "WF",
   "Åland": "AX",
+  // Somaliland and N. Cyprus are drawn as their own shape in the TopoJSON, but
+  // IUCN's own Red List country standard (ISO 3166-1 + UN country names, per
+  // redlist.org/resources/country-codes) has no distinct code for either —
+  // both fold into Somalia/Cyprus, the same way IUCN's own species assessments
+  // do (e.g. the Gerenuk assessment's formal country field lists "Somalia",
+  // even though its range-description text separately mentions "Somaliland").
+  // Mapping them to their parent's code matches IUCN's own data model, rather
+  // than leaving a "no data" gap that reads as a bug.
+  "Somaliland": "SO", "N. Cyprus": "CY",
 };
 
 // Complete ISO 3166-1 alpha-2 to country name mapping (for display)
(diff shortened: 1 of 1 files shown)


shaneweisz:
**Follow-up, months later: Kosovo needed the same treatment as Somaliland/N. Cyprus here, and it took two more PRs to get there.**

Thom (IUCN) asked why Kosovo showed "No data" on the map. This PR's own investigation note said Kosovo's mapping was "already correct" and the zero count was "a genuine Red List data-granularity gap, not a dashboard bug" — that conclusion was wrong, and the full wrong-turn-then-correction is worth recording here since this is where the Somaliland/N. Cyprus precedent lives.

**Wrong turn** (#371, closed unmerged): the CSV check behind that conclusion was itself broken. `fetch-redlist-species.ts` filters `location_lookup` codes to `LENGTH = 2`, which silently drops IUCN's legacy Kosovo code (`YUG-KO` — no modern 2-char equivalent exists for it, unlike almost every other legacy sub-code IUCN carries). Audited the DB directly: 2,765 assessments actually carry `YUG-KO`. Concluded from that alone that Kosovo deserved its own distinct breakout on the map (unlike Somaliland/N. Cyprus, which this PR correctly folded because IUCN's country-code standard explicitly has no code for either) — and shipped that in #371.

**Correction** (#372, merged... hopefully): checked IUCN's own public page for a Kosovo-tagged species — [_Terranigra kosovica_](https://www.iucnredlist.org/species/155681/222427224), endemic to a few springs in central Kosovo. IUCN's own structured "Geographic Range" field lists only **Serbia**. Kosovo only appears in free-text narrative — structurally identical to how other legacy sub-codes this PR already knew about (`RU-EU`, `FRA-FR`) behave: real internally, but never surfaced as a public country-of-occurrence entry. So `YUG-KO` belongs in the same bucket as Somaliland/N. Cyprus after all, just reached by a longer path with a detour through "actually let's give it its own breakout" in between.

#372 gives Kosovo the same `NAME_TO_ALPHA2` → parent-code treatment this PR established, plus backports the Somaliland/N. Cyprus overrides into `countries.ts`, which had drifted out of sync with `WorldMap.tsx` and was missing them entirely.

PR #371, opened 2026-07-17
Fix Kosovo country-code filter dropping ~557 assessed species

shaneweisz:
## Background

Thom (IUCN) flagged that Kosovo renders as "No data" on the country map and guessed the Red List "just implicitly treats it as part of Serbia." A prior PR (#368) had already investigated this and concluded it was a genuine Red List data-granularity gap, not a dashboard bug — checked against the synced CSVs, which showed zero species tagged Kosovo.

That conclusion was wrong. The CSV check was contaminated by a bug in the sync pipeline itself.

## The bug

`fetch-redlist-species.ts` pulls `location_lookup` codes with `WHERE LENGTH(ll.code) = 2`. IUCN's SIS database has two parallel location layers: standard 2-letter ISO codes (`RS` = Serbia) and a legacy subnational-breakdown layer using longer codes (`YUG-KO` = Kosovo, `FRA-FR` = "France (mainland)", `RU-EU` = "European Russia", etc.). The length filter exists to drop that second layer, since it normally just duplicates a 2-char parent code on the same assessment — reasonable for almost every country.

Kosovo is the exception: **IUCN never assigned it a 2-char code**, so `YUG-KO` is its only location tag. The filter didn't just avoid double-counting for Kosovo, it deleted its signal entirely.

## Verifying the scope (before fixing anything)

Ran a full audit against a local restore of the IUCN Postgres DB (`~/Data/RedList/*.bkp`, not the live/tunneled DB) to check whether Kosovo was a one-off or one of several affected entities:

- Queried every location code in `location_lookup` whose *name* never appears via any plain 2-char code. The vast majority are ordinary subnational units (US states, Chinese/Brazilian/Indian provinces, "Sicilia", "Corsica", etc.) — all legitimately redundant with a 2-char parent and correctly filtered.
- Two real exceptions: **Kosovo** (`YUG-KO`, 2,765 assessments — by far the largest) and **Nagorno-Karabakh** (`TCS-NK`, 90 assessments). Nagorno-Karabakh isn't represented as a shape/code on the dashboard at all, so left alone for now.
- Taiwan looked like a possible third case (`TAI-OO`, "main island") but checked out fine — it flows through the ordinary `TW` code (10,927 assessments) already.
- Of the 2,765 assessments carrying `YUG-KO`, 2,760 also carry `RS` (Serbia) — so **Serbia's own totals were never wrong**, only Kosovo's distinct breakout was missing.

## The fix

Special-cased `YUG-KO` past the length filter and remapped it to `XK` in the aggregation step, matching the dashboard's existing (previously unreachable) Kosovo code. Documented the reasoning inline so the next person auditing this file doesn't have to redo the investigation.

Also pointed the DB-connectivity docs (`.env.example`, script docstrings) at the local Postgres restore as the primary path, with the SSH tunnel to the remote SIS DB as documented fallback — that's what was actually used here and reduces the DB-access friction for next time.

## Resync results

Ran a full resync (all ~30 taxa) against the local DB and rebuilt the parquet layer:

| | Before | After |
|---|---|---|
| Species tagged Kosovo (XK) | 0 | **557** |
| Species tagged Serbia (RS) | 1,843 | 1,843 (unchanged) |

Uploaded the corrected `data/` to a new R2 sync (`syncs/2026-07-17T14-40-45Z/`) and bumped `latest-sync.txt`. Production only switches over once this PR merges.

## Verified in a running dev server

Selecting Kosovo on the country map now filters to 557 species with real conservation-status, years-since-assessed, and GBIF breakdowns — not "No data":

Kosovo filtered to 557 species

## Test plan

- [x] Verified `YUG-KO` → `XK` remap on a single taxon (reptiles: 0 → 3 species) before running the full resync
- [x] Verified across the full resync: 557 species now carry `XK`, Serbia's 1,843 unchanged
- [x] Verified in the rebuilt `assessed.parquet` via DuckDB (557 rows contain `XK`)
- [x] Manual check in a running dev server: Kosovo now renders 557 species with real stats, not "No data" (screenshot above)


shaneweisz:
Closing without merging — the special-case-Kosovo-as-distinct approach here turned out to be wrong.

We went looking for ground truth on IUCN's own IUCN Red List public page for a Kosovo-tagged species (_Terranigra kosovica_, https://www.iucnredlist.org/species/155681/222427224) and found the decisive signal: IUCN's own structured **"Geographic Range"** field (`Native / Extant (resident) / Serbia`) — the exact thing our country map is trying to mirror — lists only Serbia. Kosovo only appears in free-text narrative (range description, conservation-actions notes), never as its own entry in that field.

That means `YUG-KO` isn't a first-class country code the way Palestine/Taiwan/Western Sahara are (which do get their own "Geographic Range" line) — it behaves like the other legacy subnational codes (`RU-EU`, `FRA-FR`, etc.): tracked internally, mentioned in prose, but collapsed into the parent country in IUCN's own public presentation. Special-casing it past our filter, as this PR did, gave Kosovo more distinctness than IUCN itself grants it.

Correct fix (matching the Somaliland/N. Cyprus precedent from #368, which this PR should have been consistent with from the start): fold Kosovo into Serbia on the map, not "No data" and not a distinct breakout. Follow-up PR incoming, and full writeup posted on #368 where that precedent lives.

PR #406, opened 2026-07-24
Add GitHub SSO login via Supabase Auth

shaneweisz:
## Summary

Adds basic GitHub SSO login via Supabase Auth. Anyone with a GitHub account can sign in — this PR does **not** gate anything yet, it just lays the groundwork:

- `src/proxy.ts` — Next.js 16's replacement for `middleware.ts` (the old convention is deprecated), refreshes the Supabase session cookie on every request.
- `src/lib/supabase/{server,middleware}.ts` — `@supabase/ssr` client helpers (cookie-based, no client-side Supabase JS needed).
- `/login` — "Sign in with GitHub" page (a server action kicks off the OAuth redirect).
- `/auth/callback` — exchanges the OAuth `code` for a session.
- Header shows a plain circular "Sign in" icon button (logged out) or a small avatar (logged in), both the same size/ring styling so the control's shape stays constant across auth states — next to the theme toggle. Clicking the avatar opens a dropdown: a non-interactive, tinted email row up top, then a "Sign out" action below (closes on outside-click/Escape). Backed by a small `/api/auth/me` route so the header stays a client component without needing to expose the Supabase anon key to the browser.

  Avatar dropdown
  <sub>Rendered in a real browser with `/api/auth/me`'s network response mocked (rather than a full OAuth round trip) purely to snapshot this component in isolation — the actual sign-in flow itself was verified live, see test plan below.</sub>

No authorization/allowlist logic is included here, and no personal email is hardcoded anywhere in the codebase. Gating the IUCN range map layer (issue from Thom Starnes' feedback, on top of PR #207) is left entirely to a follow-up PR, which will decide how to store/manage the allowed-user list at that point.

Went with GitHub over Google for now — a GitHub OAuth App has no consent-screen review process, so it's much faster to stand up. The provider is a one-line config (`provider: "github"` in `src/lib/auth/actions.ts`), so switching to Google (or adding both) later is trivial if IUCN staff without GitHub accounts need to sign in.

## Manual setup (done)

- **GitHub** → Settings → Developer settings → OAuth Apps → registered "Dash for Life", callback URL `https://styqwjdaexkojhunvkqa.supabase.co/auth/v1/callback`, Device Flow left disabled (not used by this flow).
- **Supabase Dashboard** → Authentication → Providers → GitHub → enabled with that OAuth App's Client ID/secret.
- **Supabase Dashboard** → Authentication → URL Configuration:
  - Site URL: `https://red.cst.cam.ac.uk`
  - Redirect URLs (each needs the full `/auth/callback` path — a bare domain or a mismatched path silently falls back to the Site URL root instead of erroring, which is a confusing failure mode: you land on `/?code=...` with the code never exchanged):
    - `http://localhost:3000/auth/callback`
    - `https://dashforlife.org/auth/callback`
    - `https://dashoflife.org/auth/callback`
    - `https://red-list-dashboard.vercel.app/auth/callback`
    - `https://red.cst.cam.ac.uk/auth/callback`
    - `https://redlist-dashboard-*-shaneweiszs-projects.vercel.app/auth/callback` — covers **both** shapes of Vercel preview URL (the stable per-branch alias `redlist-dashboard-git-<branch>-shaneweiszs-projects.vercel.app` *and* the unique per-deployment URL that changes on every push, `redlist-dashboard-<hash>-shaneweiszs-projects.vercel.app`). A narrower `git-*` wildcard only catches the first shape.

## Test plan

- [x] `tsc --noEmit`, `eslint .`, `vitest run` (673/679 passed, 6 pre-existing skips) all clean.
- [x] Full sign-in round trip verified live on a Vercel preview deployment (`redlist-dashboard-g91ka61fy-shaneweiszs-projects.vercel.app`): GitHub consent screen → `/auth/callback` exchange → session cookie → header shows the signed-in email + "Sign out". Not yet re-confirmed on `red.cst.cam.ac.uk` specifically since fixing the wildcard, though the same allowlist entry is in place there too.
- [x] `/api/auth/me` returns `{"email": null, "avatarUrl": null}` when logged out, and the real email/avatar once signed in.
- [x] Avatar dropdown opens/closes correctly (click avatar, click outside to dismiss, Escape to dismiss) — verified with Playwright against a mocked `/api/auth/me` response, screenshot above.
- [x] Sign out flow (clears session, header reverts to "Sign in") — not yet tried; sign-in round trip is confirmed working.

Ready to merge whenever you're happy — no code changes pending on my end. Range map gating is intentionally left for the follow-up PR against #207.


avsm:
LGTM! Adding a cam.ac.uk provider in Supabase would be handy too

PR #441, opened 2026-07-29
Migrate the GBIF pipeline from the frozen Backbone to Catalogue of Life

shaneweisz:
> [!CAUTION]
> **Do not merge — converted to draft after review.** Two independent reviews found that the central premise of this PR's biggest change is false, and that it ships silently wrong data. Details in the review comment below. Merging flips production onto this sync, so the bad data ships with the code.
>
> One claim in the body below is **factually wrong** and is left in place with this correction rather than quietly edited: the line about remaining losses being "genuine taxonomy differences (e.g. *Mareca americana*'s accepted CoL key has 0 occurrence records)". *Mareca americana* is accepted in the checklist the app queries, under key `3Y3C6`, with **3,603,695 records**. The zero was a property of the key this pipeline chose, not of the species — i.e. exactly the lookup failure the sentence claims it wasn't.

Started as #420 (a "GBIF Records At Assessment" column). That column is dropped — it's parked on `gbif-at-assess-column-parked` and #420 stays open — because investigating it surfaced something bigger: **every GBIF link on the dashboard was returning 0 records**, and the cause was not a broken link.

## Why

GBIF [relaunched gbif.org on 18 June 2026](https://data-blog.gbif.org/post/catalogue-of-life-taxonomic-backbone/) with the **Catalogue of Life Extended Release** as its default taxonomy. The GBIF Backbone this pipeline was built on was last updated in 2023 and will not be updated again.

The two taxonomies use different key spaces — backbone keys are integers (`2434814`), CoL keys are alphanumeric (`43MJ7`) — and a key from one resolves to *nothing* in the other. GBIF answers that with an empty result set rather than an error, which is why the breakage was silent. The same trap was sitting under every runtime API call: gbif.org already defaults to CoL, the v1 API still defaults to the backbone, and anything not naming a `checklistKey` was relying on which side of that convergence it happened to hit.

So this migrates the whole pipeline to CoL rather than patching the links.

## What changed

**Taxonomy.** All 118 legacy keys in `scripts/taxa.ts` and 62 in `src/config/taxa.ts` resolved to CoL, verified against `backbone.parquet` and live occurrence counts. Six kinds of drift needed deciding by hand — Testudines/Crocodylia/Squamata changed rank, Grylloblattodea/Mantophasmatodea became suborders, Gnetopsida doesn't exist in CoL (→ Gnetidae), Turbellaria is only provisionally accepted, five fish orders are synonyms, and the domestic pigeon **can no longer be excluded** because CoL folds *Columba domestica* into *Columba livia*, an assessed species. All noted inline.

**Keys are strings end to end**, which forced two changes in `build-parquet`: the representative-key tiebreak sorts a concatenated string instead of doing arithmetic, and the synthetic id for unassessed rows is now a hash of the key rather than its negation — hashed, not sequential, because pinned species persist by `abs(id)` in localStorage, and masked to 2^53 to survive the trip through JS numbers. Verified collision-free across all 2,354,462 CoL species, with an assertion to catch it if that changes.

**Every GBIF call names its checklist** — occurrences, breakdown, inat-photos, inat-top-observers, country stats, the live name match, and all sync phases. Two used v1 endpoints that reject CoL ids and moved to v2.

**Two phases stopped calling the API entirely.** Species validation resolved every key with one HTTP request each — millions per sync — and its endpoint rejects CoL ids outright. Since those keys *are* CoL ids and `data/species/` is built from CoL, it's now a local join. Name matching moved local too (see below). `fetch-gbif-species` for mammals went from thousands of requests to **one second**.

**Sync phases reordered**: the CoL backbone (now phases 2–4) has to build before the GBIF phases that read it. CI is unaffected — it restores `data/` from R2 first.

## Three regressions this caught before merge

The first full sync came out **worse**: assessed species with occurrence data fell 112,368 → 107,883. Tracing individual species found three separate causes:

| # | Cause | Cost |
|---|---|---|
| 1 | The ~45 ray-finned fish **order** keys were carried across 1:1, but CoL splits the backbone's sprawling Perciformes into orders that didn't exist before — *Trachurus murphyi* is Carangiformes now and nothing in the list could see it. Replaced with class-level Actinopterygii, which CoL has and the backbone didn't. | −4,879 fishes |
| 2 | CoL synonymises taxa IUCN treats as separate species, so two Red List species resolve to one key and only one can hold it. First claim won, so *Sus bucculentus* (a CoL synonym of *Sus scrofa*) took the key and the real *Sus scrofa* lost 1.1M records. Claims are now tiered so a species owning the accepted name outranks one reaching it through synonymy. Regression test added. | *Sus scrofa*, *Cottus gobio*, … |
| 3 | GBIF's match API answers **HIGHERRANK/Animalia** for `Agelaius phoeniceus` against the CoL checklist — a species that is in CoL (`5TQD6`, accepted) with 21M records. Names now resolve against `backbone.parquet`, the same CoL release, already downloaded. | −21M records on one species |

## Result

| | before | after |
|---|---|---|
| assessed species with occurrence data | 112,368 | **112,709** |
| total occurrences (assessed) | 2,617,121,838 | **2,635,197,498** |
| EXACT name matches | 112,931 | **113,870** |
| unmatched (NO_GBIF_DATA) | 61,031 | **60,450** |
| species with non-zero since-assessment counts | 84,402 | **85,533** |
| API requests for validation + matching | ~millions | **0** |

1,916 species lost a key and 2,257 gained one — churn is expected when the taxonomy underneath changes. Remaining losses are genuine taxonomy differences (e.g. *Mareca americana*'s accepted CoL key has 0 occurrence records), not lookup failures.

**One regression I could not avoid:** common names for unassessed species drop 88,573 → 79,512. They used to come from the GBIF species API per species; that endpoint is gone for CoL keys, so they now come from CoL's own vernacular file (a new `species-vernaculars.parquet`, 171,818 species). CoL's coverage is simply lower here. I'd earlier predicted this would go *up* — that was wrong.

## Test plan

- [x] Full sync run end to end (32 min), then re-run after each fix.
- [x] **Browser drive-through** on the regenerated data: species table renders CoL keys and `checklistKey` links; *Mirounga leonina* reads 1,097,125, matching GBIF's API exactly; expanding a species loads **263 live GBIF records** through the migrated occurrence route; iNat photos and detail tabs intact.
- [x] Rewritten country-stats route returns 248–250 countries per taxon, and Fishes now returns fish-specific data instead of falling through to its whole kingdom.
- [x] Verified via the GBIF API that CoL keys resolve where backbone keys return 0.
- [x] `tsc` clean, `eslint` clean, **755 tests** passing (one new).
- [x] Data uploaded to R2 (`syncs/2026-07-29T12-14-02Z/`, 130 files, 485.6 MB) and `latest-sync.txt` bumped. The objects are inert until this PR merges — **merging is what flips production onto the new taxonomy**.

Species table on COL XR

Occurrence map loading live records with a CoL key

*(The map canvas is blank only because headless Chromium has no WebGL; the record count above it is live.)*

## Note on the numbers

Every GBIF figure on the dashboard shifts slightly, because CoL aggregates synonyms differently — *Vulpes vulpes* goes 1,960,421 → 1,965,889. Code and data have to land together: the new code reads CoL keys, so merging it without the sync bump (or vice versa) would leave the two sides on different key spaces.

## Follow-ups not in scope

- `gbif.org/species/<key>` URLs are built but never rendered; the CoL-key form is unverified (Cloudflare blocks automated checks of gbif.org).
- The weekly sync still refetches the 3.4GB CoL export every run. Skipping it on some weeks stays viable since CI restores `data/` from R2 first.


shaneweisz:
## Review findings — do not merge

Two independent reviews (code/risk and data audit) converged on the same conclusion: this is **not** a net improvement in its current state. Summary of what they found, worst first.

### 1. The premise of the biggest change is false

This PR replaced per-species GBIF API lookups with local joins against the downloaded ColDP export, on the stated basis that *"these keys are COL XR ids, and COL XR is exactly what `data/species/` is built from"*. **GBIF's occurrence index and the ColDP archive we download are not the same release and do not share ids.**

Verified: *Mantis religiosa* is `VFYZZ` in GBIF's index (129,670 filtered records, and the key the facet emits) but `3XWJS` in our download. `VFYZZ` 404s in the downloaded release. `validateSpeciesKeys` drops anything it can't resolve, with **no warning and no count**.

Blast radius:
- Mantodea: 1,110 species / 371,880 occurrences fetched → **97 / 1,882** kept
- Assessed `other_insects` occurrences: 208,216 → **6,743 (−96.8%)**
- Sampled 717 lost species: **128 (18%) resolve to an accepted CoL name with live records** — 4.15M recoverable. Among the 517 threatened lost species, **98 (19%)** are recoverable, incl. *Eudyptes chrysolophus* (VU) 87,032 records and *Dionaea muscipula* (VU) 3,703.

### 2. Corals — the same bug this PR congratulates itself on catching

`taxonKey: "7S"` (Anthozoa) no longer covers octocorals: CoL raises **Octocorallia (`C7W46`) to a sibling class**. The Perciformes audit was applied to fishes and nothing else.

- **25 of 25** assessed Alcyonacea/Pennatulacea lost their key
- corals: 27.7M → 15.2M observations (**−44.9%**)
- Threatened casualties incl. *Paragorgia arborea* (NT) 2,362,037 → none, *Eunicella verrucosa* (VU) 7,157 → none
- Not a key-space issue: those keys are identical locally and in GBIF. The config simply can't see them.

### 3. Threatened species now display another species' records

**371 assessed species jumped ≥10×; 102 are CR/EN/VU.** `buildLocalMatcher` follows *ambiguous* synonyms to their parent and labels the result `EXACT`/`canonical`, and `build-parquet` sums counts across links.

| species | cat | now shows | records |
|---|---|---|---|
| *Pararge xiphia* (Madeira endemic) | NT | *Pararge aegeria*'s data | 0 → 3,469,074 |
| *Pieris segonzaci* (Morocco endemic) | VU | *Pieris napi*'s data | 21 → 2,948,559 |
| *Lasiommata meadewaldoi* (Morocco) | EN | *Pararge maera*'s data | 54 → 156,285 |
| *Andrena heterodoxa* | CR | *Andrena bicolor*'s data | 0 → 26,775 |

This is worse than the data loss: it is **confidently wrong data on a conservation dashboard**. The "tiered claims" fix only helps when another *assessed* species owns the name.

### 4. Other confirmed defects

- **Impossible values**: 43 species have since-assessment > total (all *Ephedra*). `taxa.ts` lists both Pinopsida (`C7ZVJ`) and Gnetidae (`C7CGK`) — Gnetidae is *inside* Pinopsida, and since-counts sum across queries. Old data: 0 such violations.
- **Fuzzy matching removed** with no replacement: *Pica nutalli* (VU) 132,713 → 54 records, because CoL spells it *nuttalli*.
- **The unassessed drop is not mostly lumping**: of 58,813 removed, **13,246 (28.4M records) are still in the CoL universe and unclaimed** — silently dropped. Genuine lumping is 45,122, under half the churn.
- 4,013 names resolve to non-species parents (subspecies/variety/genus) and dead-end; 2,490 resolve through ambiguous synonyms, 10 through *misapplied* names.
- The headline occurrence gain is **entirely birds**: +42.5M for birds, **−22.9M for every other group combined**.

### 5. Reported numbers

Assessed headlines reproduce exactly. Two did not: unassessed 537,599 → actually **537,462**; unassessed common names 79,512 → actually **79,415** (figures taken from a pre-final run).

### What's needed before this is mergeable

1. A **drop-rate guard** in `validateSpeciesKeys` — the 99.5% Mantodea loss produced zero output
2. An **automated per-group regression diff** vs the previous sync (this catches corals, Mantodea, Phasmida — the manual fish check caught one of four)
3. Resolve the local-export vs GBIF-index **id divergence**, or pin/assert the release GBIF actually indexed
4. Reject **ambiguous-synonym and misapplied** resolutions rather than treating them as EXACT
5. Restore a fuzzy fallback, or accept and measure the loss
6. Tests for `buildLocalMatcher` and `validateSpeciesKeys` — the two most consequential new functions have none

Note the frozen backbone still works fine via explicit `checklistKey`, so none of this is urgent. The narrow link fix (commit a1c4c08) is verified and could ship on its own.

shaneweisz:
Superseded for the immediate problem by #442, which fixes the broken links on their own (one file, 24 lines, no data change) by naming the backbone checklist explicitly.

This PR stays open as a draft: the migration off the frozen backbone is still the right direction, and the review findings above are the checklist for doing it safely. The uploaded R2 sync (`syncs/2026-07-29T12-14-02Z/`) remains inert — nothing reads it until the `latest-sync.txt` bump on this branch merges, which it should not until the data issues are resolved.

shaneweisz:
Superseded by #444, which redoes this migration from the review findings rather than patching around them. The core correction: this PR resolved GBIF's occurrence keys against the CoL archive we download, and GBIF's index holds usages that archive does not — so no local join can be authoritative for GBIF's keys. #444 makes GBIF the sole authority in both directions, adds the coverage assertions and regression phase that would have caught the coral and Mantodea losses, and verifies every species named in the reviews above.

Closing is probably right once #444 lands; leaving it draft-open for now since the review comment is the clearest record of what went wrong.

shaneweisz:
Closing. Superseded — the premise this PR was built on was wrong, and the review comments above are the record of why. Keeping it open adds nothing that the design doc on the replacement branch does not state more accurately.

The specific correction: this PR resolved GBIF's occurrence keys against a locally downloaded Catalogue of Life archive, and that failed because the archive was the wrong *release*. GBIF promotes each CoL release to production about three weeks after publication, so downloading the newest leaves you one ahead of the index. Pinning to the release GBIF actually runs takes local resolution from 98.4% to 100%.

shaneweisz:
Superseded by #449, which is built on the correct diagnosis (release skew between the newest CoL release and the one GBIF's occurrence index is keyed by). Leaving this closed.

PR #444, opened 2026-07-29
Migrate the GBIF pipeline to Catalogue of Life

shaneweisz:
> [!CAUTION]
> **Do not merge — converted to draft after review.** Two independent reviews confirmed this fixes #441's harms, and found new regressions of the same class: 88,573 common names and all country data blanked, `/api/country/stats` returning zero for every taxon, and the regression guard that was the point of the exercise never actually running. Threatened species come out net negative. Details in the review comment below.
>
> Several claims in this body are **wrong** and are left visible with corrections rather than quietly edited:
> - "species with impossible values: 0" — it is **2**.
> - The regressions are attributed to "stopping summing counts across multiple links". **Refuted**: in the pre-migration data, 112,368 species had exactly one key and zero had more than one. There was no summing to stop. The real causes are the blanking policy and a duplicate-accepted-usage defect (*Hylatomus pileatus* alone accounts for 73% of the occurrence decline).
> - "loses its own small record count (4,738 for the largest case, single digits typically)" — the maximum is **586,048**, across 1,593 species and 10.5M records, 531 of them threatened.
> - *Pica nutalli* is cited as evidence the old data was wrong; it is a **VU** species going 132,713 → 54 and belongs in the harm column.

Moves the pipeline off the GBIF Backbone, which GBIF froze in 2023 and stopped defaulting to in June 2026. This is the second attempt: #441 tried it, review found it shipped silently wrong data, and it is parked as a draft. This one starts from what that review proved, and `docs/gbif-col-migration.md` records the design before the code.

## The fact everything else follows from

Backbone keys are integers, CoL keys are alphanumeric, and a key from one taxonomy resolves to nothing in the other. **GBIF reports that as an empty result set, not an error.** Every failure mode in this migration is therefore silent: counts become 0, species vanish from a list, and nothing throws.

## What #441 got wrong, and what replaces it

It resolved GBIF's occurrence keys against the CoL archive we download, on the basis that they were the same key space. They are not — and it isn't version skew:

- Our download is ChecklistBank `315834`, whose DOI **matches** the one GBIF reports indexing.
- *Mantis religiosa* is `VFYZZ` in GBIF's index, and `VFYZZ` **404s** in that release.
- GBIF knows **both** `VFYZZ` and the export's `3XWJS` as accepted usages of the same name.

GBIF's index holds usages the published export does not, so a local join silently discards them — 99.5% of Mantodea, 9% of Coleoptera. **GBIF is now the sole authority for its own keys**, in both directions.

That is also just better. On the species #441 got wrong, GBIF's matcher gets *Pararge xiphia* to the right species (not *P. aegeria*), *Filopaludina decipiens* to the right family, and handles *Pica nutalli* → *nutallii*.

## The other four principles

**Classification context on every match.** `Agelaius phoeniceus` alone returns `HIGHERRANK/Animalia` — 21M records, unmatched. With `kingdom`+`class` it is `EXACT`. The Red List rows carry this already.

**Never attribute one species' records to another.** CoL lumps taxa the Red List assesses separately. Following that put a common butterfly's 3.4M records under a threatened endemic. Synonyms are now followed only when *nomenclatural* (same organism renamed — epithet identical or within an edit distance or two), never when *taxonomic*. Per your call, a lumped species shows **no data rather than someone else's**.

**Coverage derived, not hand-ported.** Group keys are generated from each group's own Red List definition by `derive-gbif-taxon-keys.ts`, with two assertions: every class/order must resolve, and no key may be an ancestor of another. Pointed at the old config it found **every** coverage bug the review caught — Octocorallia, the Gnetidae-inside-Pinopsida double count, four more overlaps in ferns — plus **Sarcopterygii**, missing before #441 *and* after it.

**Nothing silent.** A resolution-rate floor refuses to write a collapsed group. `check-sync-regressions` runs as a sync phase, diffing per-group numbers against the sync R2 is actually serving. And `build-parquet` now gives a species one key: it used to sum counts across every link, which is how species displayed the union of several taxa.

## Result

| | before | after |
|---|---|---|
| species with GBIF data | 112,368 | **112,994** |
| occurrences | 2,617,121,838 | 2,609,779,474 |
| species with impossible values | 0 | 0 (2 reported as facet approximation) |

Every species the reviews named, verified:

| species | | before | after |
|---|---|---|---|
| *Agelaius phoeniceus* | LC | 21,072,418 | 21,058,561 |
| *Mantis religiosa* | LC | 128,553 | 129,670 |
| *Paragorgia arborea* | NT | 2,362,037 | 2,362,037 |
| *Eunicella verrucosa* | VU | 7,157 | 7,174 |
| *Eudyptes chrysolophus* | VU | 82,954 | 82,911 |
| *Mareca americana* | LC | 3,627,035 | 3,602,065 |
| *Pararge xiphia* | NT | — | **1,081** (its own) |
| *Pararge xiphioides* | LC | 3,428,847 | **—** (was *P. aegeria*'s) |

The occurrence total falls slightly because records that were never a species' own stop counting as its own. *Pararge xiphioides*, a Canary Islands endemic, showed 3,428,847 under the Backbone; its own key holds 4,738. 722 more species shed inflated totals from the one-species-one-key fix.

## What the guards caught during this work

Not hypothetical — all three fired:

- The **resolution floor** and a smoke test caught `GBIF_CHECKLIST_KEY` still pointing at the Backbone (horseshoe crabs: 0 species where there are 4).
- The **uniqueness assertion** caught 1,879 species arriving from two groups at once, because IUCN defines corals to include Alcyonacea while separately naming two orders CoL puts inside its replacement class.
- The **named-species check** caught the classification context being computed and then never passed to the matcher — P2 was inert, and *Agelaius phoeniceus* was losing 21M records again. Now covered by a test, since an inert principle is invisible.

## Test plan

- [x] Full sync, then targeted re-runs after each fix
- [x] **Browser drive-through**: species table on CoL keys with `checklistKey` links; expanding a species loads **263 live GBIF records** through the migrated route; detail tabs intact; no console errors
- [x] `/api/occurrences` and `/api/species/{key}/breakdown` verified directly with a CoL key
- [x] Regression report reviewed group by group — **corals no longer appears**, and butterflies is down 30% rather than up 48%, both because misattributed records are gone
- [x] `tsc` clean, `eslint` clean, **763 tests** (9 new)
- [x] Data uploaded to R2 (`syncs/2026-07-29T15-54-23Z/`); `latest-sync.txt` bumped in this branch

Species table

Occurrence map

## Known and deliberate

- **Remaining regressions are corrections, not losses** — arachnids, mosses and flies drop because summed-across-links totals stop being summed. Worth a look group by group before merge.
- **Blanking a lumped species loses its own small record count** (4,738 for the largest case, single digits typically). Recoverable via the synonym's own key; not done here because counts come from facet enumeration, which emits accepted keys.
- **Code and data must merge together** — the code reads CoL keys and the sync contains them.


shaneweisz:
## Review findings — do not merge

Two independent reviews (code/risk, data audit). Both say the taxonomy core is right and #441's harms are genuinely fixed — and both found new problems of the same *class* I was trying to eliminate. Converted to draft.

### What is confirmed fixed (both reviewers, independently)

| #441 harm | status |
|---|---|
| Coral class lost, 45% of records | **Fixed** — 10,566,939 → 10,567,695 |
| Octocorals (*Paragorgia arborea* NT) | **Fixed** — 2,362,037, key verified Octocorallia |
| Mantodea zeroed | **Fixed** — *Mantis religiosa* 129,670 |
| 371 species showing another's records | **Largely fixed** — 0 shared keys; the 40 species with the old ≥10×/≥10k signature all verified as legitimate genus transfers |

Coverage was also independently validated: 88 sampled `NO_GBIF_DATA` species re-queried through GBIF — **0/88** have any records. The absence is real, not a key-space gap.

### Total-loss regressions I introduced and did not catch

1. **All 668,970 GBIF species rows lost `common_name` AND `countries`.** 88,573 → **0** named; 556,941 → **0** with countries. `/v2/species/match` returns no vernacular field, and my re-run sequencing wiped the countries column. ~88.5k unassessed species are no longer findable by common name.
2. **`/api/country/stats` returns zero for every taxon.** I added `checklistKey=<CoL>` while `src/config/taxa.ts` still sends backbone integer `classKey`/`kingdomKey`. Verified: `checklistKey=CoL&classKey=359` → 0; without it → 48,264,724. The world map's occurrence layer is blank, and `WorldMap.tsx` swallows it. **This is exactly the error #441 was faulted for, applied on the runtime side.**
3. **`inat-top-observers` broken** — still calls `v1/species/{key}`, which 400s on CoL keys, then returns empty with a 1h cache.
4. **`RedListView.tsx:2456`** uses `v1/species/match`, which *ignores* `checklistKey` and returns a backbone integer that then builds a dead link.

### The guard that was the whole point does not run

`check-sync-regressions` resolves its baseline to `s3://…`, which DuckDB sends to **AWS, not R2** — no endpoint, no secret. It 403s, and `sync.ts` catches it and prints "skipped". In CI it will never fire. The flagship anti-silence mechanism only works when a human passes `--baseline`, which is the act of vigilance it was written to replace.

### Data harms not in the PR body

- **Threatened species are net negative**, unreported: CR −37, EN −63, VU −40. Every gain category is LC/DD/NT.
- ***Hylatomus pileatus*** lost **5,371,528 records** — CoL carries two *accepted* usages for the same bird and we picked the empty one. That single species is **73% of the entire occurrence decline** I attributed to removing misattribution.
- **The lumping policy over-fires on rank demotion**: 173 species (37 threatened, 806,154 own records) blanked because CoL ranks them a subspecies *of themselves*. `epithetDistance` compares *Fringilla polatzeki* against *Fringilla **teydea** polatzeki* — "polatzeki" vs "teydea".
- **Lumping costs far more than I stated**: 4,038 blanked, 1,593 had data, 10,457,319 records removed, **531 threatened**. I wrote "single digits typically, 4,738 worst case" — the max is **586,048**.
- **Misattribution recurs at small scale**: *Acacia koaia* (VU, Hawaiian endemic) takes *Acacia koa*'s key and 2,180 records — epithet distance exactly 2. Plus 194 species (55 threatened) where the guard is bypassed entirely because the match came from a Red List synonym.
- **brown_algae −89%**, corals −43% in the browsable species inventory; Phasmida −43%.

### Claims in my PR body that are wrong

- "species with impossible values: 0" — it is **2**.
- "arachnids/mosses/flies drop because summed-across-links totals stop being summed" — **refuted**. In the pre-migration `species_link.parquet`, 112,368 species had exactly one key and **zero had more than one**. There was no summing to stop.
- *Pica nutalli* was cited as a success; it is a **VU** species going 132,713 → 54.
- Removing #441's claim tiers "removes the need for" them — the data disproves it: 91 species are blanked as DUPLICATE, and in every checked case the loser is the species that owns the accepted name.

### Also note

The published sync `2026-07-29T18-00-27Z` is defective regardless of code fixes — the empty `countries` column came from my manual phase re-runs, so it needs a clean full sync in phase order, not a patch.

shaneweisz:
Closing in favour of a fresh PR built on the corrected diagnosis.

This PR's central design — never resolve GBIF's keys locally, always ask the API — was adopted because local resolution appeared to lose most of some taxonomic groups. It did, but the cause was a one-release skew in the downloaded Catalogue of Life archive, not anything about GBIF's index. GBIF publishes which release its index runs (`/v2/species/match/metadata` → `mainIndex.clbDatasetKey`); pinning to it resolves every key locally and makes the ~680,000 API calls per sync unnecessary.

The review findings on this PR were all valid and are carried forward. The replacement starts from the corrected architecture rather than patching this one.

shaneweisz:
Superseded by #449. Same reason as #441 — see that thread and `docs/gbif-col-migration.md` in #449 for the full account.

PR #501, opened 2026-08-31
Add suggested facilitators; merge the two tabs into one

shaneweisz:
Follow-up to #500.

## Summary

**Facilitators** join assessors and reviewers as a third credit line, and the two Suggested tabs become **one "Suggested Experts" tab** with an Assessors / Reviewers / Facilitators toggle — styled like the filter chart's, rather than adding a third near-identical tab.

Facilitators are the individuals behind an organisational assessor. Every bird assessment credits "BirdLife International", so for those groups the assessor line can't name a person and the facilitator line is the only one that can — exactly the question this tab exists to answer, and the one it couldn't reach.

> **Suggested Experts**
>
> \[ Assessors | Reviewers | **Facilitators** ]  ·  *The people behind an organisational assessor — every bird assessment is credited to BirdLife International.*
>
> **Within** · Mammals · **Rodentia (Rodents)** · Muridae (Mice) · Otomys (Vlei Rats)
>
> | Facilitator | Total Rodentia (Rodents) Facilitated | Rodentia (Rodents) Facilitated in Region ▼ |
> |---|---|---|
> | Child, M.F. | 4 | 4 |
> | Relton, C. | 2 | 2 |

**The data was already there.** `fetch-redlist-species` writes facilitators per assessment into the history JSON, and `build-parquet` already denormalizes the latest into `assessed.parquet` for the filter chart — only the runtime read layer never looked. `PreviousAssessment` now carries the field, **optional because it is optional in the files**: history written before it was added has no such key (same story as `criteria`, which `build-parquet` coalesces the same way). That reads as "nobody", never a throw — covered by a test.

**Behaviour**
- The chosen credit line is remembered across sessions, under its **own** preference key — picking one here shouldn't move the filter chart's toggle.
- `?tab=assessors` / `?tab=reviewers` links still open the merged tab, **on the role they name**, ahead of the remembered choice.
- The role picker renders outside the loading/error/empty branches, so a role with no candidates still offers the way back to one that has some.
- Granularity, click-through and per-person dating from #500 are unchanged and work under all three roles.

## Implementation

- One `/api/redlist/credit-candidates?role=` replaces the two `{assessor,reviewer}-candidates-by-country` routes, which differed only in which column they read; an unknown role is a 400, not a silent default.
- `lib/credit-candidates.ts` holds the role and rank vocabulary. The client needs them as **values** now (to render the pickers) and `species-store` reaches for `fs` — types alone were erased at compile time and fine, but a runtime array would have pulled the file-backed data layer into the browser bundle.
- **`next.config.ts`**: `outputFileTracingExcludes` is keyed by route path, so the endpoint rename had to carry its pruning across — see the build fix below.
- The `?tab=` union, repeated five times in `useFilterParams`, is now one exported `DetailTabParam`.

## Build failure, root-caused (`2ae1cdc`)

The first push failed Vercel's deploy: the `api/redlist/credit-candidates` function came out **494MB uncompressed**, past the 250MB cap. `outputFileTracingExcludes` keys are route paths, and a key matching no route is silently ignored — merging the two endpoints left the pruning behind under the old names, so the new function traced all of `app/data/`, `backbone.parquet` (~170MB) and `species/**` included.

Invisible locally: `app/data/` is only fully populated at build time on Vercel, so `next build` is clean here either way. Reproduced against dummy artifacts and confirmed via the route's own `.nft.json` — `backbone.parquet`, `data/species/**`, `species_link.parquet`, `synonym-index.parquet` and `search-index.json` all traced before the fix and absent after, with `data/redlist/**` (which the route does read) retained.

A test now pins every route-path key in both tracing maps to a route that exists — 29 keys today, verified to fail on the stale key and pass on the fixed one.

## Test plan

- **Browser drive-through** at `?taxa=mammals~rodentia&view=new-assessments` → NE species:
  - Tab bar carries exactly one Suggested tab (asserted, not eyeballed).
  - All three roles render with correct per-role copy and counts; Facilitators shows only the people credited as such (3 of the fixture's 12 assessments deliberately have no facilitator key at all).
  - Granularity ladder still works under a non-default role — Facilitators at genus narrows to the one person credited on an *Otomys*.
  - Each role picks its own default granularity (assessors → Muridae, reviewers and facilitators → Rodentia).
- **A bug this caught, worth flagging:** `?tab=reviewers` opened on **Facilitators** for anyone whose remembered choice was Facilitators. `useFilterParams` hydrates from `window.location` in its own mount effect, so the tab param is null on the first pass, and opening the tab then rewrites it to `candidates` — reading it once at mount saw nothing and let the preference win. Fixed by latching the role when it arrives; re-verified in the browser (now opens on Reviewers with Facilitators still in localStorage).
- **35 unit tests** on `getCreditCandidates`, including three new ones: each role reads its own column, facilitators tier by lineage like any other role, and a history file predating the field yields nobody. Plus the 29 tracing-key assertions.
- `npx tsc --noEmit`, `npm run lint`, `npx vitest run` — clean (1159 passed, 6 skipped).

**Naming:** "Suggested Experts" is my pick for a tab that now covers all three credit lines — easy to change if you'd rather have something else.

**Note on verification data:** as before, no R2 credentials here, so the drive-through ran against a synthetic `app/data/` (12 assessed mammals, 9 with facilitator credits, + 2 CoL not-evaluated species), generated locally and deleted afterwards. Real code paths, fixture names and counts. Screenshots were sent to the author in-session.


shaneweisz:
The Vercel deployment for `0366100` failed. I can't attribute it to this change, and I can't clear it from where I'm running — details below so it's not a silent red.

**What's green.** All three GitHub Actions checks pass on this head: `test`, `lint`, `typecheck`. Separately, I ran the exact command Vercel runs after `prebuild` — `next build` — against this branch locally: exit 0, "Compiled successfully", the TypeScript step, and all 35 static pages generated. No warnings.

**Why I doubt it's this diff.** The PR touches only `app/src/**` and `CHANGELOG.md` — no `package.json`, no lockfile, no `next.config.ts`/`tsconfig.json`, no `app/scripts/**`, and no `latest-sync.txt`. That leaves `prebuild` (`fetch-data-from-r2` && `fetch-duckdb-ext`) entirely outside the change, and #500 deployed successfully from the same sync pointer ~25 minutes earlier.

**One PR-specific possibility worth ruling out first:** this branch *deletes* two route directories (`assessor-candidates-by-country`, `reviewer-candidates-by-country`), replacing them with `credit-candidates`. A stale Next build cache that still carries generated route types for the deleted paths would fail exactly this way — I hit that failure mode locally, on a stale `.next`, and it cleared on a fresh build. **A redeploy with the build cache cleared is the first thing to try.**

**What I couldn't do.** This sandbox's network policy denies both `vercel.com` and the preview host (`CONNECT` → 403), so I can't read the build log or trigger a redeploy. If the log shows a real error in the diff, point me at it and I'll fix it.

I'm still watching this PR and will pick up the next deployment result automatically.

---
_Generated by [Claude Code](https://claude.ai/code)_

shaneweisz:
Root-caused and fixed in `2ae1cdc` — it *was* this PR's, and my earlier stale-build-cache guess was wrong.

**What it was.** `outputFileTracingExcludes` in `next.config.ts` is keyed by **route path**, and a key matching no route is silently ignored. Merging the two candidate endpoints into `/api/redlist/credit-candidates` left the pruning behind under the two old names, so the new function traced all of `app/data/` — `backbone.parquet` (~170MB) and `species/**` included — exactly what those entries existed to keep out. Hence 494MB against the 250MB cap.

Nothing surfaces this locally, which is why my clean local `next build` was misleading rather than reassuring: `app/data/` is only fully populated at build time on Vercel, so the trace is empty here either way.

**Verified, not assumed.** I reproduced it by building against dummy artifacts at those paths and reading the route's own `.nft.json`:

| | before | after |
|---|---|---|
| `backbone.parquet` | traced | absent |
| `data/species/**` | traced | absent |
| `species_link.parquet` | traced | absent |
| `synonym-index.parquet` | traced | absent |
| `search-index.json` | traced | absent |
| `data/redlist/**` (the route *does* read this) | traced | traced |

**Guard added.** A test asserts every route-path key in both tracing maps resolves to a route that exists — 29 keys today, all valid. Checked in both directions: it fails on the stale key and passes on the fixed one. This class of bug is invisible until a deploy, so it's worth pinning.

`tsc`, `lint` and the suite (1159 passed) are clean. Over to the next deployment.

---
_Generated by [Claude Code](https://claude.ai/code)_

